PR #145, opened 2023-02-09
Add option for unlimited pod configuration

### LINKED ISSUE #139: Add a config parameter to turn on/off CPU & memory quotas
Currently, @sk1tter and @leokondrashov have to modify the code. Need to add a proper parameter to the config file of the loader.

leokondrashov:
Signed-off-by: Leonid Kondrashov <leo.kondrashov@gmail.com>

## Summary

Add option for unlimited pods in driver configuration. Now unlimited pods may be enabled by `"container_unlimited"` as `"YAMLSelector"` in `driverConfig.json`. Resolves #139.

Add new .yaml file with limits disabled, add a case in switch inside loader to pick necessary file.

N/A

N/A

### DIFF
--- a/.github/configs/wordlist.txt
+++ b/.github/configs/wordlist.txt
@@ -138,6 +138,7 @@ Containerd
 containerPort
 coolDownTime
 Costa
+CPULimit
 CPUs
 cpu
 CRC
@@ -632,6 +633,7 @@ ustiugov
 Ustiugov
 utils
 UUID
+vCPU
 vhive
 vHive
 vHive's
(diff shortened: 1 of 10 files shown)


leokondrashov:
@cvetkovic Updated the PR with different way to set the CPU limits. It is hard to set no limits because of the format (for that we need to remove lines from yaml completely), but the limit of whole CPU should be the same, since we have a limit on concurrency of 1. I've set the whole CPU limit as default to remove the difference with other deployment methods (neither of AWS and OpenWhisk are using these limits at all).

cvetkovic (approved):

leokondrashov:
Your review looks very weird, like it is for the old version of it. I suppose, I can ignore the comments from Feb 15th. I think they were just stored for all this time in an unpublished review.

cvetkovic:
Looks like something was from the previous review attempt. I'll merge the PR as it looks good to me.

PR #234, opened 2023-07-12
Add fake k8s nvidia plugin

gaow0007:
## Summary

Add one option to install a fake nvidia-plugin which emulates GPU resource allocation. 

* Use the fake k8s-device-plugin  [container https://github.com/yzs981130/k8s-device-plugin/tree/fake_new ](https://github.com/eth-easl/loader/blob/gpu-support/config/nvidia-device-plugin.yml)
* After enable "DEPLOY_GPU" as true. Users would get   a new resource kind `nvidia.com/gpu` after running `kubectl describe nodes | grep nvidia ` on any deployed nodes. 

* The original implementation of fake k8s-device-plugin https://github.com/yzs981130/k8s-device-plugin/tree/fake_new

* 

### DIFF
--- a/config/nvidia-device-plugin.yml
+++ b/config/nvidia-device-plugin.yml
@@ -1,0 +1,62 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


ustiugov:
@gaow0007

ustiugov:
please re-open if needed

PR #254, opened 2023-09-05
Revert concurrency target to default Knative value

leokondrashov:
## Summary

Revert to default knative values for target concurrency (100% -> 70%). This would mean that autoscaler would set function scale so estimated utilization of it would be approximately 70%.

Full utilization of concurrency made system prone to fast scale ups due to variations in load. This change would make the system more stable and closer to what is expected from production system.

Remove setting it in the function configuration, better to reuse default one set on the global level.

* N/A

* N/A


leokondrashov:
@cvetkovic, what's you view on that?

cvetkovic:
> @cvetkovic, what's you view on that?

I and @Lightxyz think that you can proceed with merging. Just rebase and rerun unit tests.

cvetkovic (approved):

ustiugov:
@cvetkovic @leokondrashov please don't use merge commits, please update by rebasing

leokondrashov:
Actually, this merge commit didn't make it into the main branch, so history is still clean

Issue #248, opened 2023-08-30
Github actions fail silently


leokondrashov:
## Version & Branch

E2e tests on `main` branch.

## Behaviour

In recent test (https://github.com/eth-easl/loader/actions/runs/5998242750/job/16266139553#step:9:18), loader fails to run but test is considered successful.

leokondrashov:
@cvetkovic, can you please take a look?

cvetkovic:
Good cache. Could this solve the problem?
https://github.com/knative/serving/issues/11388#issuecomment-845787934

leokondrashov:
> https://github.com/knative/serving/issues/11388#issuecomment-845787934

I'm not quite sure I understand what you mean by that.

Actually, I have an idea that it might happening after introduction of separate loader node.
But the issue is also that this failure is silent.

cvetkovic:
You can try executing this command in the workflow before the step that fails. I think it's just a wrong Knative config. Probably the defaults changed during version upgrades.

leokondrashov:
No, actually, we didn't include nodeSelector for single-node cluster. Labels should work differently in that case.

PR #266, opened 2023-09-19
Fix end-to-end test setup

leokondrashov:
## Summary

End to end tests are failing after introduction of labels (#136). This fixes part of #248.

* Enable NodeSelector in testbed and label the node to accommodate the function.

* N/A

* N/A


leokondrashov:
@cvetkovic, do we need a 45-minute-long e2e test? That seems excessive for me.

cvetkovic (dismissed):

leokondrashov:
No, it uses cmd/config.json with ExperimentDuration set to 45, that's definitely the actual running time.

cvetkovic:
Can you have a look at the trace? I think the trace execution lasts until there are scheduled invocations, and that experiment duration is overridden in that case. Regardless of that, we should change `cmd/config.json` and set it to a minute or two.

leokondrashov:
It indeed was running for 45 minutes because there were invocations for much longer than 45 minutes. I've updated the config.

PR #267, opened 2023-09-24
Feature: Integrate with AWSLambda via serverless.com framework

LeeXuanHua:
## Summary

Added basic integration with AWS via serverless.com framework (zip deployment approach).

- Added Lambda code & Go binary under server/trace-func-go/aws/*

* 

* 

### DIFF
--- a/server/trace-func-go/aws/trace_func.go
+++ b/server/trace-func-go/aws/trace_func.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambda"
+	"github.com/vhive-serverless/loader/pkg/workload/standard"
+	"time"
+)
+
+// Response is of type APIGatewayProxyResponse since we're leveraging the
+// AWS Lambda Proxy Request functionality (default behavior)
+//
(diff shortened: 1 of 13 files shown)


cmd/config.json
```diff
@@ -1,7 +1,7 @@
 {
   "Seed": 42,
 
-  "Platform": "Knative",
+  "Platform": "AWSLambda",
```
leokondrashov, line 4: Don't change the default value here.

server/trace-func-go/aws/trace_func.go
```diff
@@ -0,0 +1,114 @@
...
+		tmp = C.SQRTSD(C.double(10))
+	}
+	return tmp
+}
+
+func busySpin(runtimeMilli uint32) {
+	totalIterations := iterationsMultiplier * int(runtimeMilli)
+
+	for i := 0; i < totalIterations; i++ {
+		takeSqrts()
```
leokondrashov, line 47: Have you tried to reuse the code from the `pkg/workload`? Since you are compiling it that shouldn't be a problem to import this module as it is done in `server/trace-func-go/trace_func.go`.

leokondrashov (changes requested):

leokondrashov:
And on top of that, apply linters to the code, fix typos and sign-off the commit. If there is no actual typo but the test fails, add it to wordlist.

leokondrashov (approved):

PR #282, opened 2023-10-04
Support for Dirigent

cvetkovic:


leokondrashov:
Please provide details bout the PR, what's Dirigent? How it operates? Does it need to be installed somehow?

PR #284, opened 2023-10-11
Support for Dirigent cluster manager

cvetkovic:
Dirigent is a cluster manager developed at ETH Zurich that is soon going to be published. In this PR, we add support for it to the loader.

### DIFF
--- a/pkg/config/parser_test.go
+++ b/pkg/config/parser_test.go
@@ -47,17 +47,17 @@ func TestConfigParser(t *testing.T) {
 	if config.Seed != 42 ||
 		config.YAMLSelector != "container" ||
 		config.EndpointPort != 80 ||
-		!strings.HasPrefix(config.TracePath, "data/traces") ||
+		!strings.HasPrefix(config.TracePath, "data/traces/example") ||
 		!strings.HasPrefix(config.OutputPathPrefix, "data/out/experiment") ||
 		config.Granularity != "minute" ||
-		config.IATDistribution != "exponential" ||
-		config.ExperimentDuration != 2 ||
+		config.IATDistribution != "equidistant" ||
+		config.ExperimentDuration != 5 ||
 		config.WarmupDuration != 0 ||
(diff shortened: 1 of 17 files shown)


pkg/driver/deployment.go
```diff
@@ -46,14 +50,50 @@ var (
...
+		deployDirigent(functions[i])
+	}
+}
+
+func deployDirigent(function *common.Function) {
+	payload := url.Values{
+		"name":                {function.Name},
+		"image":               {"docker.io/cvetkovic/dirigent_empty_function:latest"},
+		"port_forwarding":     {"80", "tcp"},
+		"scaling_upper_bound": {"1"},
```
Lightxyz, line 70: Scaling upper and lower bounds should be based on some config

Lightxyz (changes requested):

pkg/config/parser_test.go
```diff
@@ -47,17 +47,17 @@ func TestConfigParser(t *testing.T) {
 	if config.Seed != 42 ||
 		config.YAMLSelector != "container" ||
 		config.EndpointPort != 80 ||
-		!strings.HasPrefix(config.TracePath, "data/traces") ||
+		!strings.HasPrefix(config.TracePath, "data/traces/example") ||
 		!strings.HasPrefix(config.OutputPathPrefix, "data/out/experiment") ||
 		config.Granularity != "minute" ||
-		config.IATDistribution != "exponential" ||
-		config.ExperimentDuration != 2 ||
+		config.IATDistribution != "equidistant" ||
```
leokondrashov, line 53: May be we should fixate the configs that are used for tests and move them to separate folder, so we won't need to update these tests that often.

BTW, why do we need to update the default config? Is there something wrong?

Lightxyz (approved):

cvetkovic:
Let's postpone merging until the tutorial on Monday is done.

PR #310, opened 2023-11-07
Adapting API for RPS mode

cvetkovic:
Changes:
- Support for RPS mode. Direct integration into loader through separate IAT and runtime spec generation. The user can choose a target RPS, with a cold/warm start ratio, and cooldown period. Regular trace mode remains intact.
- Specification is not a matrix anymore, but an array.
- Improved error handling for Dirigent

Please be careful with reviewing/testing this PR as it affects the specification generator and trace/function driver.


pkg/generator/specification.go
```diff
@@ -126,43 +126,44 @@ func (s *SpecificationGenerator) generateIATPerGranularity(numberOfInvocations i
...
+	var perMinuteCount []int
 	var nonScaledDuration []float64
 
 	numberOfMinutes := len(invocationsPerMinute)
 	for i := 0; i < numberOfMinutes; i++ {
 		minuteIAT, duration := s.generateIATPerGranularity(invocationsPerMinute[i], iatDistribution, shiftIAT, granularity)
 
-		IAT = append(IAT, minuteIAT)
+		IAT = append(IAT, minuteIAT...)
+		perMinuteCount = append(perMinuteCount, len(minuteIAT))
```
leokondrashov, line 141: Why do you have separate array here? Shouldn't it be just the array that we already got from the trace?

pkg/generator/rps.go
```diff
@@ -0,0 +1,142 @@
...
+	"math/rand"
+)
+
+func generateFunctionByRPS(experimentDuration int, rpsTarget float64) (common.IATArray, []int) {
+	iat := 1000000.0 / float64(rpsTarget) // μs
+
+	var iatResult []float64
+	var countResult []int
+
+	duration := 0.0 // ms
```
leokondrashov, line 17: Is it ms or us? Please clarify in comments

pkg/generator/rps.go
```diff
@@ -0,0 +1,142 @@
...
+		iatResult = append(iatResult, iat)
+		duration += iat
+		currentCount++
+
+		if int(duration)/60_000_000 != currentMinute {
+			countResult = append(countResult, currentCount)
+
+			currentMinute++
+			currentCount = 0
+		}
```
leokondrashov, line 33: Didn't get that. Is it executed almost always? currentMinute is incremented by one, duration - by iat, they never will result in false condition. Except the case of iat being the full minute, then it becomes completely weird.

pkg/generator/rps.go
```diff
@@ -0,0 +1,142 @@
...
+
+	duration := 0.0 // ms
+	totalExperimentDurationMs := float64(experimentDuration * 60_000_000.0)
+
+	currentMinute := 0
+	currentCount := 0
+
+	for duration < totalExperimentDurationMs {
+		iatResult = append(iatResult, iat)
+		duration += iat
```
leokondrashov, line 25: So, for RPS mode there is only equidistant IAT with minute granularity? Why do we need this mode if we can use output of tools/trace_synthesizer with any set of IAT types and granularity?

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,376 @@
...
+		expectedIAT        common.IATArray
+		expectedCount      []int
+	}{
+		{
+			testName:           "2min_1rps",
+			experimentDuration: 2,
+			rpsTarget:          1,
+			expectedIAT: []float64{
+				// minute 1
+				1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000,
```
leokondrashov, line 23: This will be incompatible with the one generated with equidistant IAT in the trace mode: driver sleeps for the IAT in the minute and then invokes, and no invocation after last IAT in minute, since it is there just to align it to minute (or second) duration. Here, it won't generate expected number of invocations, but one less.

leokondrashov (changes requested):

pkg/generator/specification.go
cvetkovic: It ended up not being the same array. I noticed a wired behavior of the specification generator since introduction of these changes. For example, if we say we want to create 5 invocations within a minute, we get an array of length 6. The elements are 0, 12_000_000, 12_000_000, 12_000_000, 12_000_000, 12_000_000. I think this is caused by line 122.

image

pkg/driver/trace_driver_test.go
```diff
@@ -261,33 +266,39 @@ func TestDriverBackgroundProcesses(t *testing.T) {
...
+			expectedInvocations: 10,
 		},
 		{
-			testName:          "without_warmup_second_granularity",
-			withWarmup:        false,
-			secondGranularity: true,
+			testName:            "without_warmup_second_granularity",
+			withWarmup:          false,
+			secondGranularity:   true,
+			expectedInvocations: 6,
```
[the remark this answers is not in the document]
cvetkovic: The number of invocations that are generated by the spec. generator is 6, however, the driver loop executed only 5, and the one that's left ended up spilling into the next minute and was skipped. As I changed the invidiaul function driver a bit, the invocation don't spill anymore and actually end up being executed.

IMO, the specification generator should be fixed along with this.

docs/configuration.md
```diff
@@ -1,32 +1,50 @@
...
+| RpsCooldownSeconds           | int       | > 0                                                                                              | 10                  | RPS-mode target load                                                   |
+| RpsImage                     | string    | "trace" or "empty"                                                                               | "empty"             | RPS-mode function image                                                |
+| RpsRuntimeMs                 | int       | > 0                                                                                              | 10                  | RPS-mode function runtime                                              |
+| RpsMemoryMB                  | int       | > 0                                                                                              | 2048                | RPS-mode function memory footprint                                     |
+| RpsIterationMultiplier       | int       | > 0                                                                                              | 80                  | RPS-mode SQRT iterations                                               |
```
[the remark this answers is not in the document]
cvetkovic: RPS mode is applicable to any system. As we don't read from the trace, we need to specify these 4 parameters. What alternative approach would you suggest?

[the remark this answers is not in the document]
leokondrashov: For knative deployment we have this kind stuff set in the yaml file. Maybe use similar approach for Dirigent? This would remove this parameter from the loader configuration and reduce confusion about the applicability of this parameter.

Another option is to make this parameter as a replacement for ITERATIONS_MULTIPLIER that we have in the yamls, this would be applicable to both modes and different platforms.

pkg/generator/specification.go
leokondrashov: I've made this changes to have the same format of IATs for with and without shifted invocations to have the driver logic the same independent of IAT that we've chosen. Everytime, except empty minutes, it creates one more number than there are invocations. First one being the interval between beginning of the minute and first invocation, all other invocations happen after delay stated in next elements of the array. Last one would just align the sum to 60s without invocation after.

Since we are flattening the array, maybe we should merge the last element of previous minute with first of the next one. After that, we won't need this additional element for each minute. But that would change the logic of skipping the minute in driver. And there wouldn't be possible the sleeps to the end of the minute in the driver, minute counting should be careful.

pkg/generator/rps.go
cvetkovic: No. The purpose of this is to count the number of invocations within a minute, i.e. to generate the `countResult` array.

pkg/generator/rps.go
cvetkovic: Because it's not that trivial to create a trace that can trigger some cold/warm start ratios. Knative is not responsive enough, so a workaround must be used.

pkg/generator/rps_test.go
cvetkovic: I don't think this is relevant for RPS mode. The purpose of this mode is just to fire invocations for X minutes. The first and the last couple of minutes should discarded because of warmup and the issue you mention.

pkg/generator/rps.go
leokondrashov: Ah, I see, that's integer division. I still propose to be clearer with the inequality signs and replace it with <. The result should be the same.

pkg/generator/rps.go
leokondrashov: But you are creating this trace inside the loader itself. Dumping the counted invocations for each minute and replaying it will give you the same result.

Didn't get the point about knative not being responsive.

PR #325, opened 2023-11-27
Finer reference trace

JooyoungPark73:
## Summary

Add the finer number of sampled functions for the step of 10 functions between 10-200 functions.
Add suggestion of number of functions to replay relative to the number of cores (5 functions per core, 80 for 16-core D430)
Also, fix recent dependabot bump in python package versions creating dependency issue with Python 3.8 in vHive20 image, and gitignore to properly ignore only csv files.


docs/sampler.md
```diff
@@ -119,16 +119,18 @@ The reference traces are stored in `data/traces/reference` folder of this reposi
...
-python3 -m preprocess  -t data/azure/ -o data/reference/preprocessed_150 -s 00:09:00 -dur 150
+tar -xvzf sampled_150.tar.gz
+tar -xvzf preprocessed_150.tar.gz
 
-python3 -m sample -t data/reference/preprocessed_150 -o data/reference/sampled_150 -min 3000 -st 1000 -max 24000 -tr 16
-python3 -m sample -t data/reference/sampled_150/samples/3000 -o data/reference/sampled_150 -min 50 -st 50 -max 3000 -tr 16
+python3 -m sampler preprocess  -t data/azure/ -o data/reference/preprocessed_150 -s 00:09:00 -dur 150
+
+python3 -m sampler sample -t data/reference/preprocessed_150 -o data/reference/sampled_150 -min 3000 -st 1000 -max 24000 -tr 16
+python3 -m sampler sample -t data/reference/sampled_150/samples/3000 -o data/reference/sampled_150 -min 50 -st 50 -max 3000 -tr 16
```
leokondrashov, line 133: Add command to add the smaller traces here as well.

docs/sampler.md
```diff
@@ -119,16 +119,18 @@ The reference traces are stored in `data/traces/reference` folder of this reposi
 `preprocessed.tar.gz` contains the preprocessed traces for the original Azure trace for day 1, 09:00:00-11:30:00 (150
 minutes total).
 
-`sampled.tar.gz` contains the sampled traces for preprocessed trace from `preprocessed.tar.gz`. Sample sizes are 50-3k
-functions with step 50 and 3k-24k with step 1k.
+`sampled.tar.gz` contains the sampled traces for preprocessed trace from `preprocessed.tar.gz`. Sample sizes are 10-200 functions with step 10, 200-3k with step 50, and 3k-24k with step 1k.
```
leokondrashov, line 122: I've noticed that we have a different name for the archive (sampled_150 instead of sampled). Can you please fix it?

requirements.txt
```diff
@@ -1,7 +1,7 @@
 matplotlib==3.7.2
-numpy==1.26.1
+numpy==1.24.4
 pandas==1.3.5
-scipy==1.11.2
+scipy==1.10.1
```
leokondrashov, line 4: @cvetkovic We noticed that this bump of numpy and scipy versions make it impossible to use sampler on Ubuntu 20.04 (I think that's the reason, but might be other factor). Do you know a proper way of fixing that?

JooyoungPark73: It is due to the Python version, but still, the Python version is linked to the Ubuntu version which is annoying, so I lowered the dependency versions. The numpy 1.26.1 and scipy 1.10.1 require Python 3.9, and we don't have that as default on Ubuntu 20.04

docs/sampler.md
```diff
@@ -113,22 +113,29 @@ optional arguments:
...
+The reference traces are stored in `data/traces/reference` folder of this repository, as `preprocessed_150.tar.gz` and
+`sampled_150.tar.gz` files stored in Git LFS.
 
-`preprocessed.tar.gz` contains the preprocessed traces for the original Azure trace for day 1, 09:00:00-11:30:00 (150
+`preprocessed_150.tar.gz` contains the preprocessed traces for the original Azure trace for day 1, 09:00:00-11:30:00 (150
 minutes total).
 
-`sampled.tar.gz` contains the sampled traces for preprocessed trace from `preprocessed.tar.gz`. Sample sizes are 50-3k
-functions with step 50 and 3k-24k with step 1k.
+`sampled_150.tar.gz` contains the sampled traces for preprocessed trace from `preprocessed.tar.gz`. Sample sizes are 10-200 functions with step 10, 200-3k with step 50, and 3k-24k with step 1k.
```
leokondrashov, line 122: `preprocessed` is still here

leokondrashov (changes requested):

docs/loader.md
```diff
@@ -124,6 +124,8 @@ For to configure the workload for load generator, please refer to `docs/configur
 There are a couple of constants that should not be exposed to the users. They can be examined and changed
 in `pkg/common/constants.go`.
 
+We suggest at most 5 functions per core with SMT disabled. For example, 80 functions for 16-core node. If you increase function instance longer than that, trace replaying will start to fail.
```
ustiugov, line 127:
```suggestion
Sample sizes appropriate for performance evaluation vary depending on the platform. 
As a starting point for fine-tuning, we suggest at most 5 functions per core with SMT disabled. 
For example, 80 functions for a 16-core node. With larger sample sizes, trace replaying may lead to failures in function invocations.
```

requirements.txt
ustiugov: ok for now but we need to move to using ubuntu 22 for measurements. this is about time but we need to plan this upgrade vHive-wide.

ustiugov (changes requested):
minor things and can merge

leokondrashov (approved):

PR #341, opened 2023-12-27
Handle SIGTERM in the image

leokondrashov:
## Summary

In current implementation, created pods are staying in terminating state for the whole grace period of 300s (default value in knative). However, that is not intended behaviour by the authors (https://github.com/knative/serving/issues/13600#issuecomment-1473829791). That happened due to the fact that our image didn't received the SIGTERM and didn't handle it properly.

With this fix, duration of terminating stage is reduced significantly (to approx. 30s), which greatly reduces the memory utilization of the node resources (due to the memory released earlier by the terminating pods).

* Change the entrypoint of the image to exec form to receive the signals. Previous usage of shell form was hiding the signals completely.
* Exec form of entrypoint is incompatible with env variable substitution, so the binary was change to get parameters straight from the environment.
* Add handler for the SIGTERM and gracefully shutdown the grpc server on arrival.

* N/A

* In general, images should be used as it was before. In cases when need to replace the args, use env variables should be used now.


cvetkovic (approved):
LGTM! Have you regenerated the images?

leokondrashov:
I've regenerated the images

PR #357, opened 2024-01-23
Deploying multiple `serverless.yml` to handle large number of functions

LeeXuanHua:
## Summary

Fixed 2 main issues with regards to **AWS Lambda deployment**:
1. Multiple `serverless.yml` to handle any number of Lambda functions
**Problem:** Unable to deploy large number of AWS Lambda functions in a single `serverless.yml` file due to limitations
**Solution:** 
    1. Minimised file size of deployed serverless zip through `package.patterns` in `serverless.yml` (to avoid the [75GB Lambda Code storage limit](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html))
    2. Disabled [Lambda versioning](https://www.serverless.com/framework/docs/providers/aws/guide/functions#:~:text=Versioning%20Deployed%20Functions) through `provider.versionFunctions` in `serverless.yml`  (to minimise resources required per `.yml` file and avoid [CloudFormation 500 Resource Limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html#:~:text=AWS%20CloudFormation%20template.-,500%20resources,-To%20specify%20more))
    3. Split invocation traces into groups of 70 functions (to fit within [IAM maximum policy size of 10,240 bytes](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#:~:text=Role%20policy%20size%20can%27t%20exceed%2010%2C240%20characters.))
    4. Deploy all groups in parallel of 2 `serverless.yml` files (to avoid Out Of Memory Error while reducing deployment time)
**Note: Parallel deployment of 2 files is adjustable depending on the available memory of the compute instance

2. Added **Lambda timeout duration** to fit within **API Gateway timeout**
**Problem:** Current invocation failures could be due to many reasons such as
**(1) AWS Lambda timeout -** AWS Lambda functions automatically timeout and fail if `requestedDuration` is more than AWS Lambda's default 6s
**(2) AWS API Gateway timeout -** AWS API Gateway automatically timeout and fail if `requestedDuration` is more than AWS API Gateway's accepted 29s
**Solution:** Minimise invocation failures by explicitly limiting Lambda functions' timeout duration to 29s. Note that if `requestedDuration` > 29s, program would still run and Lambda would timeout at 29s, returning `5xx` error status code.

### DIFF
--- a/pkg/driver/deployment_awslambda.go
+++ b/pkg/driver/deployment_awslambda.go
@@ -3,6 +3,7 @@ package driver
 import (
 	log "github.com/sirupsen/logrus"
 	"github.com/vhive-serverless/loader/pkg/common"
+	"strings"
 	"sync"
 	"sync/atomic"
 )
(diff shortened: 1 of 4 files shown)


tools/trace_synthesizer/synthesizer.py
```diff
@@ -94,7 +94,7 @@ def generate(args):
             pattern = p * repetitions
 
             invArr.extend(pattern)
-        else:
+        elif mode == 2:
```
leokondrashov, line 97: Is that indended to be changed in this PR? And also, I don't see much reason to change that at all.

leokondrashov (commented):
Looks good to me as it is. But are there any tests? Both for existing stuff and new.

@ustiugov please also take a look.

leokondrashov:
@cvetkovic Can you also take a look at this?

pkg/driver/deployment_awslambda.go
```diff
@@ -3,31 +3,110 @@ package driver
...
+					}
+
+					serverless.CreateServerlessConfigFile(index)
+
+					log.Debugf("Deploying serverless-%d.yml", index)
+					// Deploy serverless functions and update the function endpoints
+					functionToURLMapping := DeployServerless(index)
+
+					if functionToURLMapping == nil {
+						log.Fatalf("Failed to deploy serverless-%d.yml", index)
```
cvetkovic, line 47: Do you need to clean up functions that have successfully been deployed? Does the user get billed for this somehow?

cvetkovic (approved):

pkg/driver/deployment_awslambda.go
leokondrashov: Idle functions are not billed, only the invocations and execution time. But the clean up should be done, remaining functions will conflict with the newly deployed.

@LeeXuanHua can you look at this? Can we do a clean up here, so we are cleaning the deployed functions even in the case of failure to deploy some of them?

leokondrashov: The point is that it is not ran in case of failure to deploy: it exits straight away on Fatalf. But for now, it happens rarely, and we will have separate PR for automatic clean-up and handling such cases.

LeeXuanHua: Apologies, you are correct

PR #358, opened 2024-01-24
Chore: Fix Missing MIT License in Files, trace_synthesizer documentation & defining explicit behaviour 

LeeXuanHua:
## Summary

Chore to add MIT license to some files, improve documentation and code of trace_synthesizer tool (explicitly define code behaviour)

* N/A

* N/A

* Using mode >= 3 for trace_synthesizer tool will lead to runtime failure instead of defaulting to mode 2

### DIFF
--- a/tools/trace_synthesizer/README.md
+++ b/tools/trace_synthesizer/README.md
@@ -47,6 +47,26 @@ optional arguments:
 ### Example
 
 ```bash
-python3 . generate -f 2 -b 10 -t 20 -s 5 -dur 3 -e 500 -mem 350 -o example -m 0
+# Normal mode is to invoke all functions at increasing rate
+# E.g. For 2 functions, starting at 10 RPS and maximum of 20 RPS with step = 5, each function will be invoked 600 times at 1st, 2nd and 3rd minute; 900 times at 4th, 5th and 6th minute; 1200 times at 7th, 8th and 9th minute
+# RPS is multiplied by 60 by default, to change this, edit `ipm = [60 * x for x in rps]` in `synthesizer.py`
+# `-dur` determines how long each RPS should continue for (in minutes) before increasing to the next RPS step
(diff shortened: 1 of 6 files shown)


tools/trace_synthesizer/README.md
```diff
@@ -47,6 +47,26 @@ optional arguments:
 ### Example
 
 ```bash
-python3 . generate -f 2 -b 10 -t 20 -s 5 -dur 3 -e 500 -mem 350 -o example -m 0
+# Normal mode is to invoke all functions at increasing rate
+# E.g. For 2 functions, starting at 10 RPS and maximum of 20 RPS with step = 5, each function will be invoked 600 times at 1st, 2nd and 3rd minute; 900 times at 3rd, 4th and 5th minute; 1200 times at 5th, 6th and 7th minute
```
leokondrashov, line 51: Numbers don't add up. We have 3rd and 5th minutes twice.

tools/trace_synthesizer/README.md
```diff
@@ -47,6 +47,26 @@ optional arguments:
...
+# RPS is multiplied by 60 by default, to change this, edit `ipm = [60 * x for x in rps]` in `synthesizer.py`
+# `-dur` determines how long each RPS should continue for (in minutes) before increasing to the next RPS step
+python3 . generate -f 2 -b 10 -t 20 -s 5 -dur 3 -e 500 -mem 350 -o example_normal -m 0
+
+# RPS Sweep mode is to spread the invocations across all functions evenly
+# E.g. Across the default 10 minute cycle (i.e. `padding=10`) and for 2 functions, each function will be invoked once every 5 minute
```
leokondrashov, line 57: Maybe better to say, functions will be invoked one by one every 5 minutes. Is that correct? Right now, it seems that all the functions are invoked every 5 minutes, and that's the same as burst mode.

leokondrashov (changes requested):

leokondrashov (approved):

Issue #329, opened 2023-12-05
HTTP mode waits for end of all invocations before writing the results


leokondrashov:
## Version & Branch

main branch, HTTP mode (used by OpenWhisk and AWS Lambda modes).

## Expected Behaviour

Write `trace` logs and output files right after getting response from server.

## Actual Behaviour

Wait for all invocations and then write both to the logs and durations file.

## Steps to Reproduce the Problem

1. Run any trace with > 1 invocation in OpenWhisk or AWS Lambda mode with `trace` verbosity. Correct response info will be printed after all invocations are finished.

## Additional Info

This was originally used in OpenWhisk mode, after that extracted to InvokeHTTP function and reused by AWS Lambda mode.

leokondrashov:
@MashaNes @cvetkovic Seems that you were the original authors of OpenWhisk mode. Is there any reason for the Wait() [here](https://github.com/vhive-serverless/invitro/blob/main/pkg/driver/http_client.go#L220)?

cvetkovic:
This is so that OpenWhisk performance is not affected by running the workload and querying the results simultaneously.

PR #360, opened 2024-01-31
Fix: Adjusted `httpInvocation()` to write results immediately on return

LeeXuanHua:
## Summary

Previously, HTTP mode would wait for the end of all invocations before writing the results to log & CSV. After this fix, all HTTP returns would be written immediately for ease of result interpretation.

Moved `AnnounceDoneExe.Wait()` out of `httpInvocation()` to prevent such behaviour from affecting all HTTP-based invocations. Kept `AnnounceDoneExe.Wait()` under `InvokeOpenWhisk()` to ensure OpenWhisk performance is not affected by running the workload and querying the results simultaneously. Refer to [Issue 329](https://github.com/vhive-serverless/invitro/issues/329).

N/A

N/A


pkg/driver/http_client.go
```diff
@@ -57,6 +57,7 @@ func InvokeOpenWhisk(function *common.Function, runtimeSpec *common.RuntimeSpeci
 	log.Tracef("(Invoke)\t %s: %d[ms], %d[MiB]", function.Name, runtimeSpec.Runtime, runtimeSpec.Memory)
 
 	success, executionRecordBase, res := httpInvocation("", function, AnnounceDoneExe, true)
+	AnnounceDoneExe.Wait() // Refer to Issue 329: https://github.com/vhive-serverless/invitro/issues/329
```
leokondrashov, line 60: Better to add an explanation for that from the comment in the issue instead. Like 'To postpone querying OpenWhisk during the experiment for performance reasons'.

leokondrashov (commented):
LGTM, just fix the comment.

PR #379, opened 2024-02-20
Kubernetes control plane high-availability mode

cvetkovic:
## Summary

* 

* 


scripts/setup/create_multinode.sh
```diff
@@ -53,6 +53,16 @@ if [ $PODS_PER_NODE -gt 1022 ]; then
     exit 1
 fi
 
+if [ "$#" -lt $CONTROL_PLANE_REPLICAS ]; then
+    echo "Not enough nodes to set up the requested number of control plane replicas."
```
leokondrashov, line 57: Need to reflect this in the docs/loader.md. I suppose, first 1, 3, or 5 nodes from the argument list will be used for control-plane.

Also, how it would play along with labeling the nodes for master/monitoring/worker?

scripts/setup/create_multinode.sh
```diff
@@ -94,20 +114,27 @@ function setup_master() {
...
-    MN_CLUSTER="pushd ~/vhive/scripts > /dev/null && ./setup_tool create_multinode_cluster ${OPERATION_MODE} && popd > /dev/null"
+    MN_CLUSTER="pushd ~/vhive/scripts > /dev/null && ./setup_tool create_multinode_cluster ${OPERATION_MODE} ${CONTROL_PLANE_REPLICAS} && popd > /dev/null"
     server_exec "$MASTER_NODE" "tmux send -t master \"$MN_CLUSTER\" ENTER"
 
     # Get the join token from k8s.
     while ! server_exec "$MASTER_NODE" "[ -e ~/vhive/scripts/masterKey.yaml ]"; do
         sleep 1
     done
 
+    MASTER_LOGIN_TOKEN=$(server_exec "$MASTER_NODE" \
```
leokondrashov, line 125: These two variables differ by addition of --control-plane --certificate-key <certificate_key>, can they be merged inton one?

leokondrashov (changes requested):
For sure, we need to add documentation for that. Also, check the labeling of the nodes to work properly.

scripts/setup/create_multinode.sh
cvetkovic: Added to `docs/loader.md`. Labeling is handled automatically by `kubeadm`. All control plane nodes get their proper role when listed with `kubectl get nodes`.

scripts/setup/create_multinode.sh
cvetkovic: Not really. MASTER_LOGIN_TOKEN is used for joining the cluster as the control plane node, whereas the other one is used to join just as a regular node. The variables can be concatenated, but I am not that familiar with `awk` so I just created two separate variables.

scripts/setup/create_multinode.sh
leokondrashov: No, I meant the label.sh that we use. It has labels for master, monitoring (which we use for load generator) and worker.

cvetkovic: This should work because all control replicas get `control-plane` tag.
image

leokondrashov: Loader node will be mislabelled in case of more than 1 master node, since in current implementation we are using the second node that we pass to the script for that. With several master nodes, we will change that behaviour, we will need to change the $LOADER_NODE_NAME to appropriate one here.

cvetkovic: Fixed. 
image

PR #383, opened 2024-02-28
"Creating a simple DAG Driver for Knative"

NotAnAddictz:
## Summary

In current implementation, single execution of ```go run cmd/loader.go --config cmd/config.json``` would create an ```individualFunctionDriver``` per function and invoke individual functions with a set frequency. 
The new implementation would create a new ```functionsDriver``` which invoke all the functions sequentially, following the invocation frequency of the first function.

* Added a new parameter ```DAGMode``` in ```cmd/config.json``` to toggle DAGMode.
* When DAGMode = true, ```functionsDriver()``` in the full list of provided functions and invokes them sequentially in a loop. 
* Added a new ```Node``` structure to hold function pointers and pointer to the next Node
* Edited parameter in ```invokeGRPC()``` to reflect invocation stats for each function.
Example Diagram (Sequential Invocation)

* (N/A)

* (N/A)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,7 +19,7 @@
 | MetricScrapingPeriodSeconds  | int       | > 0                                                                 | 15                  | Period of Prometheus metrics scrapping                                               |
 | GRPCConnectionTimeoutSeconds | int       | > 0                                                                 | 60                  | Timeout for establishing a gRPC connection                                           |
 | GRPCFunctionTimeoutSeconds   | int       | > 0                                                                 | 90                  | Maximum time given to function to execute[^4]                                        |
-
+| DAGMode             | bool      | true/false                                                          | false               | Sequential invocation of all functions one after another                                                    |
 [^1]: The second granularity feature interprets each column of the trace as a second, rather than as a minute, and
 generates IAT for each second. This feature is useful for fine-grained and precise invocation scheduling in experiments
 involving stable low load.
--- a/pkg/driver/trace_driver.go
+++ b/pkg/driver/trace_driver.go
@@ -25,16 +25,16 @@
 package driver
 
 import (
+	"container/list"
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
 	"math"
 	"os"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
-
-	"strconv"
 
 	"github.com/gocarina/gocsv"
 	log "github.com/sirupsen/logrus"
@@ -98,6 +98,15 @@ func (d *Driver) runCSVWriter(records chan interface{}, filename string, writerD
 	}
 
 	writerDone.Done()
+}
+
+func DAGCreation(functions []*common.Function) *list.List {
+	linkedList := list.New()
+	// Assigning nodes one after another
+	for _, function := range functions {
+		linkedList.PushBack(function)
+	}
+	return linkedList
 }
 
 /////////////////////////////////////////
@@ -167,9 +176,8 @@ func (d *Driver) CreateMetricsScrapper(interval time.Duration,
 /////////////////////////////////////////
 
 type InvocationMetadata struct {
-	Function              *common.Function
-	RuntimeSpecifications *common.RuntimeSpecification
-	Phase                 common.ExperimentPhase
+	RootFunction *list.List
+	Phase        common.ExperimentPhase
 
 	MinuteIndex     int
 	InvocationIndex int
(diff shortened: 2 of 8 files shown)


leokondrashov:
Please elaborate more on the operation of the proposed changes, why, what, how in the PR description. The documentation should also be in the repo docs.

Also, follow instructions from vhive [contributing guide](https://github.com/vhive-serverless/vHive/blob/main/docs/contributing_to_vhive.md) and signoff your commits.

NotAnAddictz:
Have updated the PR description, loader.md document and signed off. Do let me know if there is anything else to update.

leokondrashov:
Please be more descriptive of how it should be used. Like, you didn't mention the structure of the new file with DAG structure. Moreover, there is no example file. This is important since not everyone would dive into the code to determine the file's structure.

In addition, fix the spellcheck CI (add the word into the wordlist `.github/configs/wordlist.txt`).

docs/configuration.md
```diff
@@ -19,7 +19,7 @@
 | MetricScrapingPeriodSeconds  | int       | > 0                                                                 | 15                  | Period of Prometheus metrics scrapping                                               |
 | GRPCConnectionTimeoutSeconds | int       | > 0                                                                 | 60                  | Timeout for establishing a gRPC connection                                           |
 | GRPCFunctionTimeoutSeconds   | int       | > 0                                                                 | 90                  | Maximum time given to function to execute[^4]                                        |
-
+| DAGMode             | bool      | true/false                                                          | false               | Sequential invocation of multiple functions (only applicable for Knative platform)                                                    |
```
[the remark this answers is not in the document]
leokondrashov: Better to explain that right now, we invoke **all** the functions one after another.

pkg/driver/trace_driver.go
```diff
@@ -184,6 +193,22 @@ type InvocationMetadata struct {
 	ReadOpenWhiskMetadata *sync.Mutex
 }
 
+type InvocationMetadataDAG struct {
+	Function              []*common.Function
```
leokondrashov, line 197: This is the only field that is different from InvocationMetadata (Apart from deleted unused field). Can you reuse the original one?

leokondrashov (changes requested):
Please remove duplicate code. Also, the line
> Added new structs InvocationMetadataDAG to parse in the list of functions into individualDAGDriver()

in the description confused me, that we have some file that has the description of the DAG, while we just invoke all of them one after another. Please fix the wording.

pkg/driver/trace_driver.go
```diff
@@ -100,6 +100,27 @@ func (d *Driver) runCSVWriter(records chan interface{}, filename string, writerD
...
+		}
+		if head == nil {
+			head = node
+		} else {
+			prev.NextNode = node
+		}
+		prev = node
+		numberOfNodes++
+	}
+	return head, numberOfNodes
```
leokondrashov, line 121: numberOfNodes that is returned is just size of the functions array. I don't see a reason to count them again.

pkg/driver/trace_driver.go
```diff
@@ -203,43 +224,57 @@ func (d *Driver) invokeFunction(metadata *InvocationMetadata) {
...
-			d.Configuration.LoaderConfiguration,
-		)
-	default:
-		log.Fatal("Unsupported platform.")
+	var runtimeSpecifications *common.RuntimeSpecification
+	for {
+		if d.Configuration.LoaderConfiguration.DAGMode {
+			runtimeSpecifications = &node.Function.Specification.RuntimeSpecification[metadata.MinuteIndex][metadata.InvocationIndex]
+		} else {
+			runtimeSpecifications = metadata.RuntimeSpecifications
```
leokondrashov, line 234: Do we need to pass the RuntimeSpecification if we can get them in a general way from the function specification inside the current node? I think we shouldn't have a condition on DAGMode in invocation, we consider this method always in this mode since we are passing the linked list of functions to it.

NotAnAddictz: In normal uses, we can get them. However, for the test function ```testInvokeFunctionFromDriver``` in 
```pkg/driver/trace_driver_test.go```, they do not initialize the ```Function.Specification``` field, which contains the RuntimeSpecification. Should I do so?

**Comment by leokondrashov** [comment]
LGTM. @cvetkovic Please review as well.

File: pkg/driver/trace_driver.go [hunk 5]
```diff
@@ -100,6 +100,25 @@ func (d *Driver) runCSVWriter(records chan interface{}, filename string, writerD
...
+			NextNode: nil,
+		}
+		if head == nil {
+			head = node
+		} else {
+			prev.NextNode = node
+		}
+		prev = node
+	}
+	return head
```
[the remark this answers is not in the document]
NotAnAddictz: Yes, that is correct.

pkg/driver/trace_driver.go
```diff
@@ -203,43 +211,51 @@ func (d *Driver) invokeFunction(metadata *InvocationMetadata) {
...
-	default:
-		log.Fatal("Unsupported platform.")
-	}
-
-	record.Phase = int(metadata.Phase)
-	record.InvocationID = composeInvocationID(d.Configuration.TraceGranularity, metadata.MinuteIndex, metadata.InvocationIndex)
-
-	metadata.RecordOutputChannel <- record
+	var runtimeSpecifications *common.RuntimeSpecification
+	for node != nil {
```
cvetkovic, line 217: Are you sure this loop is going to terminate? The list is double-linked and has no end.

cvetkovic (commented):
Can you write a unit test for a trace driver with DAG mode enabled?

pkg/driver/trace_driver.go
NotAnAddictz: Yes. The list is not connected as a loop. The ```DAGCreation()``` function will add each function as the next of the previous function (next of function 0 will be function 1 ...). When ```node.Next()``` is called at the last function, it will return nil. Thus the condition ``` for node != nil ``` will be false and the for loop will terminate

NotAnAddictz: The ```DAGMode``` option is only used in ```internalRun()``` and takes in the list of functions to create the doubly linked-list. The logic of traversing through the list is done by ```invokeFunction()```. For the unit test, is it okay to simulate the initial part (creating the doubly linked list) and call ```invokeFunction()```  to emulate DAGMode being enabled? Such as in the most recent commit

cvetkovic (approved):

leokondrashov (approved):
LGTM! Thanks for the added test

PR #384, opened 2024-03-03
Fix: Use Lambda URL to bypass Gateway timeout limit

LeeXuanHua:
## Summary

**Fixed AWS API Gateway timeout limit for AWS Lambda deployment, by replacing API Gateway with Lambda function URL to generate invocation URL**

**Problem:** Previous deployment with API Gateway (with differing page paths for each Lambda function) can only handle invocations within 29 seconds due to gateway timeout. However, actual traces involve invocations longer than 29s. Although our current code filters for invocations under 1 minute, it is not sufficient to address the problem.

**Included minor fixes to improve the logging description of errors.

**Solution:** Utilise AWS' Lambda function URL feature to increase the maximum invocation duration to 15 minutes (same as Lambda execution timeout limit). 

Therefore, program can now handle any `requestedDuration` within 15 minutes and the resulting function URLs no longer share the same domain name (unlike previous approach where different Lambda functions share the same domain name due to usage of an API Gateway and only differ for the page path)

* Relies on Serverless.com framework for deployment, and that resulting function URLs are displayed in console/terminal in sequence (to map the function URL to the correct Lambda function)

* N/A

### DIFF
--- a/pkg/driver/deployment_awslambda.go
+++ b/pkg/driver/deployment_awslambda.go
@@ -3,7 +3,6 @@ package driver
 import (
 	log "github.com/sirupsen/logrus"
 	"github.com/vhive-serverless/loader/pkg/common"
-	"strings"
 	"sync"
 	"sync/atomic"
 )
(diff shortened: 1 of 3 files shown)


pkg/driver/deployment_awslambda.go
```diff
@@ -48,10 +48,7 @@ func DeployFunctionsAWSLambda(functions []*common.Function) {
 					} else {
 						atomic.AddUint64(&counter, 1)
 						for i := 0; i < len(functionGroup); i++ {
-							// Extract 0 from trace-func-0-2642643831809466437 by splitting on "-"
-							shortName := strings.Split(functionGroup[i].Name, "-")[2]
-
-							functionGroup[i].Endpoint = functionToURLMapping[shortName]
+							functionGroup[i].Endpoint = functionToURLMapping[strconv.Itoa(i)]
```
leokondrashov, line 51: How would it work with several batches of functions to deploy? Wouldn't it have the same iterator for different batches?

LeeXuanHua: As discussed, will make the edit to minimise confusion for code clarity. Thanks!

LeeXuanHua:
@cvetkovic, could you help to review the above changes, thanks!

cvetkovic (approved):

PR #398, opened 2024-03-29
Fix: AWS Golang Runtime Deprecation (zip deployment)

LeeXuanHua:
## Summary

Code deployment to AWS is failing as AWS has deprecated the `go1.x` runtime in line with the end-of-life of Amazon Linux 1. We need to migrate from `go1.x` runtime to `provided.al2023` or `provided.al2` runtime to continue with zip deployment on Lambda.

Refer to AWS announcement blog for more details: https://aws.amazon.com/fr/blogs/compute/migrating-aws-lambda-functions-from-the-go1-x-runtime-to-the-custom-runtime-on-amazon-linux-2/

As recommended in the above announcement blog:
- Updated the runtime to `provided.al2023`  for its longer support time
- Updated the handler and binary file name to `bootstrap`, and avoided nested file paths

* N/A

* N/A

### DIFF
--- a/pkg/driver/deployment_awslambda.go
+++ b/pkg/driver/deployment_awslambda.go
@@ -3,12 +3,21 @@ package driver
 import (
 	log "github.com/sirupsen/logrus"
 	"github.com/vhive-serverless/loader/pkg/common"
+	"os/exec"
 	"sync"
 	"sync/atomic"
 )
 
 func DeployFunctionsAWSLambda(functions []*common.Function) {
 	provider := "aws"
+
+	// Copy built Go binary to avoid nested directories during deployment (https://aws.amazon.com/fr/blogs/compute/migrating-aws-lambda-functions-from-the-go1-x-runtime-to-the-custom-runtime-on-amazon-linux-2/)
+	// Go binary is built in Linux by: CGO_ENABLED=1 GOARCH=amd64 GOGCCFLAGS=-m64 GOOS=linux go build -o ./server/trace-func-go/aws/bootstrap ./server/trace-func-go/aws/trace_func.go
+	copyBinaryCmd := exec.Command("cp", "./server/trace-func-go/aws/bootstrap", "./bootstrap")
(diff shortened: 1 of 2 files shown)


leokondrashov:
Do you think we need to rebuild the binary for each deployment?

@francois141 Based on our discussion, we will move to images, right? So, will image deployment supersede this PR?

I think we just need to fix the deployment method. I don't think that we need to switch to the image deployment if we can't reuse the one that is already in our ghcr. We can stop here.

LeeXuanHua:
@leokondrashov Although we can provide the default binary file (named `bootstrap`), it must be located at the project's root folder (i.e. `invitro/bootstrap`) in order to work with serverless.com framework and the new AL2023 runtime filepath requirements. Alternatively, we could place the default binary file at `invitro/server/trace-func-go/aws/bootstrap` and perform the required file operations (i.e. copying the binary file to the root folder like `invitro/bootstrap`) for successful deployment.

Currently, the code explicitly defines the required flags and OS architecture to build the binary to execute successfully in AWS's AL2023 runtime. This could be helpful subsequently if there is a need to either:
1. Change the Go binary source code
2. Change the build flags and/or instructions

That being said, we can include instructions to build customised binary in the relevant README documentations instead of specifying in the code.

Do let me know which approach is more ideal, and I can change them accordingly.

leokondrashov:
I though about deleting the binary during clean up. Can we just reuse the binary without rebuilding it each time we run experiment in the same setup?

LeeXuanHua:
@leokondrashov As discussed, we will not be continuing with this PR as we will no longer support zip deployment following this incident. Instead, we will move over to image deployment via ECR. I will proceed to close this PR.

LeeXuanHua:
@leokondrashov As discussed, we will not be continuing with this PR as we will no longer support zip deployment following this incident. Instead, we will move over to image deployment via ECR. I will proceed to close this PR.

PR #399, opened 2024-03-29
Fix: Resolved Data Race Issue caused by DAGInvocation Test

NotAnAddictz:
## Summary

The fix is to address the Data Race Issue caused by the commit 31c635c19204ee7cd4aa18867a6d1c3d11850052 .

* The Issue is caused by multiple goroutines simultaneously calling ```StartGRPCServer()```, which interacts with global variables. Since the test involves sequentially invoking a single function multiple times, ```StartGRPCServer()``` should be called once only. 

* Moved ```StartGRPCServer()``` out of for loop. 

* NA

* NA

### DIFF
--- a/pkg/driver/trace_driver_test.go
+++ b/pkg/driver/trace_driver_test.go
@@ -191,21 +191,23 @@ func TestDAGInvocation(t *testing.T) {
 	testDriver := createTestDriver()
 	var failureCountByMinute = make([]int64, testDriver.Configuration.TraceDuration)
 	list := list.New()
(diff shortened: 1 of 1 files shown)


leokondrashov:
The reason is not entirely correct. We cannot run several `StartGRPCServer()` because this module has global variables. It is okay to use the same instance of the gRPC server in this case.

@cvetkovic Do you know if the goroutines persist between tests? If they do, we should make a single gRPC server for all tests there.

leokondrashov (approved):

PR #407, opened 2024-04-13
Feature: Changed Support from Zip to Image Deployment on AWS via ECR

LeeXuanHua:
## Summary

Due to the following reasons, the zip deployment approach for AWS Lambda is no longer supported by our project. Instead, the image deployment via ECR will be adopted:

1. AWS deprecation of `go1.x` runtime (as highlighted in an earlier [PR](https://github.com/vhive-serverless/invitro/pull/398))
2. Limitations of zip deployment (primarily the maximum number of functions deployable due to [75GB code storage limit](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html))
3. Alignment to other existing image deployments

Refactored code to handle errors more gracefully.

<b>New Feature: Image Deployment</b>
* Pulls the default image (built using `./Dockerfile.trace.aws` and `./server/trace-func-go/aws/trace_func.go`) from our [public GHCR repository](https://github.com/vhive-serverless/invitro/pkgs/container/invitro_trace_function_aws)
* Creates a private AWS ECR repository using AWS CLI and upload the above image using Docker CLI (Unable to directly use [non-AWS or public image repositories](https://repost.aws/questions/QUEgw2bC3aTV2hmssdD5bYfA/how-to-use-a-public-ecr-with-aws-lambda))
* Proceeds to generate and deploy the `serverless.yml` file defining the CloudFormation stack and continue with experiments/invocations

<b>Refactor: Error Handling</b>
* Checks for external dependencies such as AWS CLI, Docker CLI, Serverless.com framework
* Checks and cleans up previously-created AWS resources, if any
  > Notably, for the rare occasion where Log Groups in CloudWatch may persist even after a clean up via `sls remove`, the log group naming has been updated to `/aws/lambda/trace-func-<count>`. This allows us to clean up obsolete log group resources through prefix filtering.

  > As a result, the number of functions per serverless.yml has decreased from <b>70</b> to <b>60</b> to fit within [IAM maximum policy size of 10,240 bytes](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#:~:text=Role%20policy%20size%20can%27t%20exceed%2010%2C240%20characters.)
* Cleans up existing AWS resources when encountering a `log.Fatalf()` error
* Failure to set up the ideal AWS environment would lead to termination of the program. In other words, majority of the responsibility to avoid deployment errors lies in the environment initialisation phase.

* AWS CLI (new)
* Docker CLI (new)
* Serverless.com framework

* Zip deployment to AWS lambda is no longer supported

### DIFF
--- a/docs/loader.md
+++ b/docs/loader.md
@@ -181,18 +181,37 @@ For instructions on how to use the loader with OpenWhisk go to `openwhisk_setup/
 
 ## Running on Cloud Using Serverless Framework
 
-Currently supported vendors: AWS
+**Currently supported vendors:** AWS
 
-Setup Instructions:
-1. Sign up for a Serverless account [here](https://app.serverless.com/)
-2. Install Serverless framework via command line `npm install -g serverless` to allow our Go code to interact with the Serverless CLI framework e.g. `sls deploy`
-3. Follow their [setup guide](https://www.serverless.com/framework/docs/getting-started) to link the respective cloud providers
+**Quick Setup for AWS Deployment:**
+1. Install the dependencies required for AWS deployment
+    ```bash
+    bash ./scripts/setup/install_aws_dependencies.sh <loader_node@IP>`
+    ```
+2. In the loader node, set AWS credentials as environment variables
+    ```bash
+    export AWS_ACCESS_KEY_ID=
+    export AWS_SECRET_ACCESS_KEY=
+    export AWS_DEFAULT_REGION=us-east-1
+    ```
+   > For more fine-grained IAM setup, please refer to Serverless [setup guide](https://www.serverless.com/framework/docs/providers/aws/guide/credentials/)
+3. In `cmd/config.json`, change `"Platform": "Knative"` to `"Platform": "AWSLambda"` (as specified in [`docs/configuration.md`](../docs/configuration.md))
+    ```bash
+    cd loader/
+    sed -i 's/"Platform": "Knative"/"Platform": "AWSLambda"/g' cmd/config.json
+    ```
+4. Start the AWS deployment experiment:
+    ```bash
+    go run cmd/loader.go --config cmd/config.json
+    ```
 
-To run the default Loader:
-1. Change the `Platform` value in `cmd/config.json` to one of those specified in `docs/configuration.md`
-2. ~~Build the Go binary in Linux: `go build -v -o ./server/trace-func-go/aws/trace_func ./server/trace-func-go/aws/trace_func.go` (For Window users, please use WSL to build the binary)~~
-3. Start the experiments: `go run cmd/loader.go --config cmd/config.json --verbosity trace`
-
+---
 Note:
-- Current deployment is via zip file where the Go binary has to be built prior to running the code. For now, the Go binary has been packaged in this repo. Otherwise, refer to Step 2 above to build your custom Go binary.
+- Current deployment is via container image.
 - Refer to [Single Execution](#single-execution) section for more details on the experiment configurations.
+- **[Strongly Recommended]** For experiments with concurrency > 10, please raise a request to increase the default AWS Lambda concurrency limit of 10 to a higher value (e.g. 1000).
+  - Go to the AWS Management Console, select the appropriate region (i.e. `us-east-1`) and search for `Service Quotas`
+  - Under `Manage Quota`, select `AWS Lambda` service and click `View quotas` (Alternatively, click [here](https://us-east-1.console.aws.amazon.com/servicequotas/home/services/lambda/quotas))
+  - Under `Quota name`, select `Concurrent executions` and click `Request increase at account level` (Alternatively, click [here](https://us-east-1.console.aws.amazon.com/servicequotas/home/services/lambda/quotas/L-B99A9384))
+  - Under `Increase quota value`, input `1000` and click `Request`
+  - Await AWS Support Team to approve the request. The request may take several days or weeks to be approved.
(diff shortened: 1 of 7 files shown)


docs/loader.md
```diff
@@ -187,12 +187,12 @@ Setup Instructions:
 1. Sign up for a Serverless account [here](https://app.serverless.com/)
 2. Install Serverless framework via command line `npm install -g serverless` to allow our Go code to interact with the Serverless CLI framework e.g. `sls deploy`
 3. Follow their [setup guide](https://www.serverless.com/framework/docs/getting-started) to link the respective cloud providers
+4. Install AWS CLI and Docker CLI (ensure that root permission is not required for `aws` and `docker` commands)
```
leokondrashov, line 190: Does it need to be done manually for each loader? If so, please make a script for automatic installation.

pkg/driver/deployment_awslambda.go
```diff
@@ -88,13 +85,153 @@ func CleanAWSLambda(functions []*common.Function) {
 		wg.Wait()
 	}
 
-	log.Debugf("Deleted %d out of %d serverless.yml files", counter, len(functionGroups))
+	if counter != uint64(len(functionGroups)) {
+		log.Errorf("Deleted %d out of %d serverless.yml files", counter, len(functionGroups))
```
leokondrashov, line 89: This case would print both "deleted x out of y" and "deleted all" lines. I guess, it should be returning after that.

pkg/driver/deployment_awslambda.go
```diff
@@ -22,46 +30,36 @@ func DeployFunctionsAWSLambda(functions []*common.Function) {
...
-
-					serverless.CreateServerlessConfigFile(index)
-
 					log.Debugf("Deploying serverless-%d.yml", index)
 					// Deploy serverless functions and update the function endpoints
 					functionToURLMapping := DeployServerless(index)
 
 					if functionToURLMapping == nil {
-						log.Fatalf("Failed to deploy serverless-%d.yml", index)
+						CleanAWSLambda(functions)                               // Clean up all deployed functions before exiting
```
leokondrashov, line 40: Go allows variable capture by reference in the lambdas. So, if it is the only reference to `functions` inside the goroutine, we can remove it from arguments, for clarity reasons.

scripts/setup/install_aws_dependencies.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
+
+#
+# MIT License
+#
+# Copyright (c) 2023 EASL and the vHive community
```
leokondrashov, line 6: No EASL toched this file. Either put your name or HySCALE. Also, bump the date.

scripts/setup/install_aws_dependencies.sh
```diff
@@ -0,0 +1,103 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+LOADER_NODE=$1
+GO_VERSION=1.20.4
```
leokondrashov, line 28: Does it compile with this version? In go.mod we have 1.21, so it shouldn't. Please bump.

scripts/setup/install_aws_dependencies.sh
```diff
@@ -0,0 +1,103 @@
...
+  *)
+    echo "Unsupported architecture $arch"
+    exit 1
+    ;;
+esac
+
+echo "Installing the dependencies for AWS deployment"
+
+# ========== Get loader ==========
+server_exec "git clone --branch=main https://github.com/vhive-serverless/invitro.git loader"
```
leokondrashov, line 52: It's not necesserily main branch. You can use the one that set in setup.cfg in the same way as create_multinode.sh uses.

leokondrashov (commented):
Looks almost done. Please fix couple minor commits.

scripts/setup/install_aws_dependencies.sh
LeeXuanHua: Copyright (c) 2024 HySCALE

Is the above ok?

scripts/setup/install_aws_dependencies.sh
```diff
@@ -0,0 +1,108 @@
...
+  *)
+    echo "Unsupported architecture $arch"
+    exit 1
+    ;;
+esac
+
+# Obtain configuration variables
+DIR=$(cd "$(dirname ${BASH_SOURCE[0]})" > /dev/null 2>&1 && pwd)
+source "$DIR/setup.cfg"
+LOADER_BRANCH=$(echo -e $LOADER_BRANCH | sed -e 's/[[:space:]]*$//')  # Remove trailing whitespace (i.e. convert main? to main)
```
leokondrashov, line 52: Do you have a specific reason to do that? I've not encountered this kind of problem while using the create_mulitnode.sh. The version is specified (by default) in quotes, it's pretty hard to accidentally put there that would have trailing symbols.

leokondrashov (approved):

PR #412, opened 2024-04-18
Feature: Added End-to-End Integration Test for Cloud Deployment (AWS)

LeeXuanHua:
## Summary

Includes an end-to-end CI/CD pipeline on GitHub Actions to test for AWS Lambda deployment.

- To maintain cloud deployment isolation, we need to ensure a **limit of 1 concurrent cloud deployment**
- This is achieved using a 3rd party GitHub Action in the marketplace which:
  1. Checks for any existing **same-named workflows**
  2. **Enqueues** the new request if the above workflows are detected
  3. Poll for their turn periodically and maintain execution order/sequence for each enqueued workflow (polling interval specified at **1 minute**; for up to **60 minutes**)
- To minimise queue time (esp. during peak periods) and compute duration, we reduced the experiment duration of cloud deployments from 5 minute to 5 seconds by changing the `granularity` in `cmd/config.json`

**Note:** We had to adopt an external dependency as the [in-built GitHub option for concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) at the job/workflow level does not provide the desired behaviour. With only a queue length of 1, when multiple workflows get queued, at most the current workflow in execution and the latest-queued workflow will be kept. All other workflows would be cancelled. (Read more [here](https://github.com/orgs/community/discussions/41518) or understand through an [example](https://github.com/orgs/community/discussions/12835)).

* [action-workflow-queue](https://github.com/ahmadnassri/action-workflow-queue)

* N/A

### DIFF
--- a/.github/workflows/e2e_aws.yaml
+++ b/.github/workflows/e2e_aws.yaml
@@ -1,0 +1,64 @@
+name: End-to-End AWS Tests
+
+on:
+ schedule:
+   - cron: "0 9 * * 1"
+ workflow_dispatch:
+ push:
+   branches: [main]
+ pull_request:
+   branches: [main]
+
+env:
+  GOOS: linux
+  GO111MODULE: on
+
+jobs:
+  test-aws:
+    name: Test AWS Cloud Deployment
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_KEY }}
+      AWS_DEFAULT_REGION: us-east-1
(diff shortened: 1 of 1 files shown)


LeeXuanHua:
@leokondrashov Could you assist to add the AWS Credentials as GitHub secret, named as `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`?

leokondrashov:
These secrets are called AWS_ACCESS_KEY and AWS_SECRET_KEY. Since the workflow comes from a branch outside the repository, we might need to manually start it if it uses secrets.

LeeXuanHua:
Will close this PR due to inability to access repository secrets from a forked branch.

Have created a branch in this upstream repository here: #414

PR #414, opened 2024-04-19
Feature: Added End-to-End Integration Test for Cloud Deployment (AWS)

LeeXuanHua:
## Summary

Includes an end-to-end CI/CD workflow on GitHub Actions to test for AWS Lambda deployment.

This workflow is triggered only upon `push` to the `main` branch, or manually from the `Actions` tab.

- As workflow is triggered only upon commit to `main` branch, PRs will not capture any failure in AWS end-to-end deployments before merging
  - Currently, workflow trigger is disabled on `pull_request` as secrets are not accessible from forked repository
  - As the code for AWS deployment is rather localised, the above workflow trigger setting will minimise any disruptions to developers
  - If required, developers can manually trigger the workflow under the `Actions` tab
- The required secrets (AWS credentials) will be verified prior to running the workflow
  - If missing / not set, workflow will fail as early as possible
- To maintain cloud deployment isolation, we need to ensure a **limit of 1 concurrent cloud deployment**
  - This is achieved using a 3rd party GitHub Action in the marketplace which:
    1. Checks for any existing **same-named workflows**
    2. **Enqueues** the new request if the above workflows are detected
    3. Poll for their turn periodically and maintain execution order/sequence for each enqueued workflow (polling interval specified at **1 minute**; for up to **60 minutes**)
  - To minimise queue time (esp. during peak periods) and compute duration, we reduced the experiment duration of AWS deployment from 5 minutes to 5 seconds by using `pkg/config/test_config_aws.json`

**Note:** We had to adopt an external dependency as the [in-built GitHub option for concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) at the job/workflow level does not provide the desired behaviour. With only a queue length of 1, when multiple workflows get queued, at most the current workflow in execution and the latest-queued workflow will be kept. All other workflows would be cancelled. (Read more [here](https://github.com/orgs/community/discussions/41518) or understand through an [example](https://github.com/orgs/community/discussions/12835)).

* [action-workflow-queue](https://github.com/ahmadnassri/action-workflow-queue)

* N/A

### DIFF
--- a/.github/workflows/e2e_aws.yaml
+++ b/.github/workflows/e2e_aws.yaml
@@ -1,0 +1,65 @@
+name: End-to-End AWS Tests
+
+on:
+ workflow_dispatch:
+ push:
+   branches: [main]
+
+env:
+  GOOS: linux
+  GO111MODULE: on
+
+jobs:
+  test-aws:
+    name: Test AWS Cloud Deployment
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_KEY }}
+      AWS_DEFAULT_REGION: "us-east-1"
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check if environment variables are set  # Fails workflow if the secrets are not set
+        # Checks for missing secrets due to branch being out of upstream repo (i.e. forked repo cannot access secrets in upstream repo)
+        run: |
+          if [[ -z "$AWS_ACCESS_KEY_ID" ]]; then
+            echo "AWS_ACCESS_KEY_ID is not set. Please check if secrets.AWS_ACCESS_KEY is in the repository. Note that forked repo cannot access secrets in upstream repo!"
+            exit 1
+          fi
+          if [[ -z "$AWS_SECRET_ACCESS_KEY" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e_aws.yaml
```diff
@@ -0,0 +1,64 @@
...
+          lfs: "true"
+
+      - name: Checkout LFS objects
+        run: git lfs checkout
+
+      # AWS CLI v2 and Docker Client & Server v24 are pre-installed
+      - name: Install Golang (Ubuntu 20.04 Cached Tool)
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22
```
leokondrashov, line 44: We are using 1.21 in the repo

.github/workflows/e2e_aws.yaml
```diff
@@ -0,0 +1,64 @@
...
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_KEY }}
+      AWS_DEFAULT_REGION: "us-east-1"
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check if environment variables are set  # Fails workflow if the secrets are not set
+        run: |
+          if [[ -z "$AWS_ACCESS_KEY_ID" ]]; then
+            echo "AWS_ACCESS_KEY_ID is not set. Please check if secrets.AWS_ACCESS_KEY is in the repository"
```
leokondrashov, line 24: make an emphasis that this is likely caused by branch being out of the upstream repo.

leokondrashov (approved):

PR #434, opened 2024-05-24
Async request driver for Dirigent

cvetkovic:


cmd/config_dirigent_dandelion_trace.json
```diff
@@ -2,6 +2,7 @@
   "Seed": 42,
 
   "Platform": "Dirigent-Dandelion",
+  "InvokeProtocol" : "http1",
```
francois141, line 5: Maybe write httpVersion and choose between 1 and 2 as integer

cmd/config_dirigent_rps.json
```diff
@@ -2,14 +2,19 @@
...
-  "RpsColdStartRatioPercentage": 50,
+  "AsyncMode": false,
+  "AsyncResponseURL": "10.0.1.3:8082",
+  "AsyncWaitToCollectMin": 1,
+
+  "RpsTarget": 10,
+  "RpsColdStartRatioPercentage": 0,
   "RpsCooldownSeconds": 10,
-  "RpsImage": "empty",
+  "RpsImage": "docker.io/cvetkovic/dirigent_empty_function:latest",
```
francois141, line 17: I think the rps image can be hardcoded in the codebase

francois141 (changes requested):

cmd/config_dirigent_dandelion_trace.json
cvetkovic: We support GRPC also, so we cannot name the parameter `httpVersion`.

francois141: Ok this makes sense. Then you can discard this comment

pkg/driver/async_processing.go
```diff
@@ -0,0 +1,119 @@
...
+	"io"
+	"math"
+	"net"
+	"net/http"
+	"strconv"
+	"sync"
+	"time"
+)
+
+func (d *Driver) writeAsyncRecordsToLog(logCh chan interface{}) {
```
francois141, line 16: I am not a huge fan of having a chan interface. Can you specify a type here? I assume the log as a homogeneous type.

pkg/driver/async_processing.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (d *Driver) getAsyncResponseData(client *http.Client, endpoint string, guid string) ([]byte, int) {
+	req, err := http.NewRequest("GET", "http://"+endpoint, bytes.NewReader([]byte(guid)))
+	if err != nil {
+		log.Errorf("Failed to retrieve Dirigent response for %s - %v", guid, err)
+		return []byte{}, 0
+	}
+
+	// TODO: set function name for load-balancing purpose
+	//req.Header.Set("function", function.Name)
```
francois141, line 97: Maybe you might want to fix this todo

pkg/driver/clients/http_client.go
```diff
@@ -37,11 +38,9 @@ type MatrixRequest struct {
 	Sets []InputSet `bson:"sets"`
 }
 
-func InvokeDirigent(function *common.Function, runtimeSpec *common.RuntimeSpecification, client *http.Client, isDandelionOptional ...bool) (bool, *mc.ExecutionRecord) {
-	isDandelion := false
-	if len(isDandelionOptional) > 0 {
-		isDandelion = true
-	}
+func InvokeDirigent(function *common.Function, runtimeSpec *common.RuntimeSpecification, client *http.Client, cfg *config.LoaderConfiguration) (bool, *mc.ExecutionRecord) {
+	isDandelion := strings.Contains(cfg.Platform, "Dandelion")
```
francois141, line 42: I insist on that. This doesn't sound the cleanest solution. Why can't you use an = operator here?

pkg/driver/clients/http_client.go
```diff
@@ -137,15 +134,16 @@ func InvokeDirigent(function *common.Function, runtimeSpec *common.RuntimeSpecif
 		return false, record
 	}
 
-	var deserializedResponse FunctionResponse
-	err = json.Unmarshal(body, &deserializedResponse)
-	if err != nil {
-		log.Warnf("Failed to deserialize Dirigent response.")
+	if cfg.AsyncMode {
+		record.AsyncResponseGUID = string(body)
```
francois141, line 138: GUID or UUID?

pkg/driver/deployment/aws_lambda.go
```diff
@@ -1,15 +1,28 @@
-package driver
+package deployment
 
 import (
 	log "github.com/sirupsen/logrus"
 	"github.com/vhive-serverless/loader/pkg/common"
 )
```
francois141, line 7:
``` go
func NewAWSLambdaDeployement() (*awsLambdaDeployer, *awsLambdaDeploymentConfiguration) {
return nil, nil // TODO Fill this
```

pkg/driver/deployment/deployer.go
```diff
@@ -0,0 +1,40 @@
...
+		configuration = knativeDeploymentConfiguration{
+			YamlPath:          cfg.YAMLPath,
+			IsPartiallyPanic:  cfg.LoaderConfiguration.IsPartiallyPanic,
+			EndpointPort:      cfg.LoaderConfiguration.EndpointPort,
+			AutoscalingMetric: cfg.LoaderConfiguration.AutoscalingMetric,
+		}
+	case "OpenWhisk", "OpenWhisk-RPS":
+		deployer = &openWhiskDeployer{}
+		configuration = openWhiskDeploymentConfiguration{}
+	case "AWSLambda", "AWSLambda-RPS":
```
francois141, line 27:
```go
return NewAWSLambdaDeployement()
```

You can then return like this and do this the same for every deployer. Then this fine can even be outside of the package.

pkg/driver/trace_driver.go
```diff
@@ -665,34 +597,14 @@ func (d *Driver) RunExperiment(skipIATGeneration bool, readIATFromFIle bool) {
...
-		DeployFunctionsOpenWhisk(d.Configuration.Functions)
-	case "AWSLambda", "AWSLambda-RPS":
-		DeployFunctionsAWSLambda(d.Configuration.Functions)
-	case "Dirigent", "Dirigent-RPS", "Dirigent-Dandelion", "Dirigent-Dandelion-RPS":
-		DeployDirigent(d.Configuration.LoaderConfiguration.DirigentControlPlaneIP, d.Configuration.Functions)
-		go scheduleFailure(d.Configuration.LoaderConfiguration)
-	default:
-		log.Fatal("Unsupported platform.")
-	}
+	deployer, configuration := deployment.CreateDeployer(d.Configuration)
```
francois141, line 600: You might have to have one single returned object. Configuration is only used in the scope of the deployer. Therefore you can create it locally only.

pkg/driver/trace_driver.go
```diff
@@ -42,44 +48,38 @@ import (
...
+		AsyncRecords: common.NewLockFreeQueue[*mc.ExecutionRecord](),
 	}
-}
 
-func (c *DriverConfiguration) WithWarmup() bool {
-	if c.LoaderConfiguration.WarmupDuration > 0 {
-		return true
-	} else {
-		return false
+	switch driverConfig.LoaderConfiguration.InvokeProtocol {
```
francois141, line 73: What about gRPC here?

francois141 (approved):

PR #460, opened 2024-07-03
RPS mode branch rebased on main

cvetkovic:
Change list:
- Support for Dirigent cluster manager - will be open-source after publication
- Support for Dandelion runtime - will be open-source after publication
- Sync/Async support for Dirigent
- Busy loop on container startup toggle
- Modular function deployment and invocation mechanism
- Code reorganization
- Failure triggering
- RPS mode and generation of trace based on specified requirements
- IO percentage busy looping


workloads/container/trace_func_go.yaml
```diff
@@ -29,7 +29,7 @@ spec:
                 - worker
                 - singlenode
       containers:
-        - image: ghcr.io/vhive-serverless/invitro_trace_function:latest
+        - image: docker.io/cvetkovic/dirigent_trace_function:latest
```
leokondrashov, line 32: What's this image? Should we use a default one? Should we rebuild it?

scripts/setup/create_multinode.sh
```diff
@@ -125,27 +130,25 @@ function setup_workers() {
...
-        server_exec $node "pushd ~/vhive/scripts > /dev/null && ./setup_tool setup_worker_kubelet ${OPERATION_MODE} && popd > /dev/null"
+
+        server_exec $node "source .profile && cd vhive; ./setup_tool setup_worker_kubelet $OPERATION_MODE"
 
         if [ "$OPERATION_MODE" = "firecracker" ]; then
             setup_vhive_firecracker_daemon $node
         fi
 
-        server_exec $node "sudo ${LOGIN_TOKEN}"
+        server_exec $node "sudo kubeadm join ${MASTER_IP}:${MASTER_PORT} --token ${MASTER_TOKEN} --discovery-token-ca-cert-hash ${MASTER_TOKEN_HASH}"
```
leokondrashov, line 140: That was exactly what was done by the old implementation. Why do we need to change it?

data/traces/example/dirigent.csv
leokondrashov, line 1: Why don't we do this for all of the platforms? Most of the parameters have a meaning for knative, for example. Knative setup would also benefit from per-function setting, for running the mixture of functions, for example.

data/traces/example/dirigent.csv
```diff
@@ -1,5 +1,5 @@
-HashFunction,Image,Port,Protocol,ScalingUpperBound,ScalingLowerBound,IterationMultiplier
-c13acdc7567b225971cef2416a3a2b03c8a4d8d154df48afe75834e2f5c59ddf,empty,80,tcp,5,0,80
-a2faad786b3c813b12ce57d349d5e62f6d0f22ceecfa86cd72a962853383b600,empty,80,tcp,5,0,80
-7dc5aeabc131669912e8c793c8925cc9928321f45f13a4af031592b4611630d7,empty,80,tcp,5,0,80
-ae8a1640fa932024f59b38a0b001808b5c64612bd60c6f3eb80ba9461ba2d091,empty,80,tcp,5,0,80
\ No newline at end of file
+HashFunction,Image,Port,Protocol,ScalingUpperBound,ScalingLowerBound,IterationMultiplier,IOPercentage
+c13acdc7567b225971cef2416a3a2b03c8a4d8d154df48afe75834e2f5c59ddf,docker.io/cvetkovic/dirigent_empty_function:latest,80,tcp,5,0,102,0
```
leokondrashov, line 2: If I understand correctly, function hash is not globally unique. We need to use all three (owner, app and func) to identify the function.

pkg/generator/specification_test.go
```diff
@@ -150,90 +147,37 @@ func TestSerialGenerateIAT(t *testing.T) {
...
-					12000000,
-					12000000,
-					12000000,
-					12000000,
-				},
-			},
-			testDistribution: false,
-		},
-		{
-			testName:        "1min_25ipm_uniform_shift",
```
leokondrashov, line 203: These tests guarantee that we didn't screw up the random generation. I agree that checking for the exact match of random number is quite weird, but we can run sanity check for the result (sum, count, etc)

cmd/config_dirigent_rps.json
```diff
@@ -0,0 +1,39 @@
...
+  "Seed": 42,
+
+  "Platform": "Dirigent-RPS",
+  "InvokeProtocol" : "http2",
+  "EndpointPort": 80,
+
+  "DirigentControlPlaneIP": "10.0.1.253:9092",
+  "BusyLoopOnSandboxStartup": false,
+
+  "AsyncMode": false,
```
leokondrashov, line 11: No reference in docs

pkg/metric/record.go
```diff
@@ -76,38 +76,10 @@ type ExecutionRecord struct {
...
-	AutoscalerStableQueue float64 `csv:"autoscaler_stable_queue"`
-	AutoscalerPanicQueue  float64 `csv:"autoscaler_panic_queue"`
-
-	SchedulingP99   float64 `csv:"scheduling_p99"`
-	SchedulingP50   float64 `csv:"scheduling_p50"`
-	E2ePlacementP99 float64 `csv:"e2e_placement_p99"`
-	E2ePlacementP50 float64 `csv:"e2e_placement_p50"`
-
-	ColdStartCount int `csv:"coldstart_count"`*/
+	AsyncResponseID     string `csv:"-"`
```
leokondrashov, line 79: As I understood, this is only applicable to Dirigent, right? Then we shouldn't add it into general structure.

pkg/metric/record.go
```diff
@@ -76,38 +76,10 @@ type ExecutionRecord struct {
...
+	TimeToSubmitMs      int64  `csv:"timeToSubmitMs"`
+	UserCodeExecutionMs int64  `csv:"userCodeExecutionMs"`
```
leokondrashov, line 81: Sounds like our actual duration field. Are they different?

pkg/driver/clients/invoker.go
```diff
@@ -0,0 +1,40 @@
...
+		} else {
+			return newHTTPInvoker(cfg)
+		}
+	case "Dirigent-Dandelion", "Dirigent-Dandelion-RPS":
+		return newHTTPInvoker(cfg)
+	case "Knative", "Knative-RPS":
+		if cfg.InvokeProtocol == "grpc" {
+			return newGRPCInvoker(cfg)
+		} else {
+			return newHTTPInvoker(cfg)
```
leokondrashov, line 31: Our image works with grpc. Do we have some other images to deploy with knative?

leokondrashov (changes requested):
This PR is enormous. I've looked only at some parts of it. For some, I couldn't give a proper review (parts focused on Dirigent and Dandelion, mostly). So if you can isolate these changes from everything else, it would be easier to sift through.

PR #526, opened 2024-10-18
Feature to trigger cluster manager component failures

cvetkovic:


pkg/driver/failure/triggers.go
```diff
@@ -0,0 +1,108 @@
...
+
+			go func(command []string, node string, t int) {
+				defer wg.Done()
+
+				finalCommand := append([]string{"ssh", "-o", "StrictHostKeyChecking=no", node}, command...)
+				invokeCommand(finalCommand, t)
+			}(command, node, t)
+		}
+
+		wg.Wait()
```
leokondrashov, line 59: This code is duplicated in the next method. Can you make it into something like `invokeCommandsOnNodes`?

pkg/driver/failure/triggers.go
```diff
@@ -0,0 +1,108 @@
...
+		case "Dirigent", "Dirigent-RPS":
+			triggerDirigentFailure(config.FailNode, config.FailComponent, config.FailAt)
+		default:
+			logrus.Errorf("No specified failure handler for given type of system.")
+		}
+	}
+}
+
+func triggerKnativeFailure(nodes string, component string, t int) {
+	time.Sleep(time.Duration(t) * time.Second)
```
leokondrashov, line 28: I would recommend moving this line into `ScheduleFailure`. This would deduplicate it and make it more obvious.

pkg/driver/failure/knative_delete_data_plane.sh
```diff
@@ -0,0 +1,13 @@
...
+kubectl delete pod $(kubectl get pods -n knative-serving -o name | cut -c 5- | grep activator | tail -n 1) -n knative-serving &
+
+# cluster-local-gateway
+kubectl delete pod $(kubectl get pods -n istio-system -o name | grep cluster-local-gateway | cut -c 5- | tail -n 1) -n istio-system &
+
+# istio-ingressgateway
+kubectl delete pod $(kubectl get pods -n istio-system -o name | grep istio-ingressgateway | cut -c 5- | tail -n 1) -n istio-system &
+
+# istiod
+kubectl delete pod $(kubectl get pods -n istio-system -o name | grep istiod | cut -c 5- | tail -n 1) -n istio-system &
```
leokondrashov, line 13: I would say istiod is the control plane. At least, based on docs, it is responsible for configuring the other istio services.

leokondrashov (commented):
LGTM, please fix couple of minor comments.

cvetkovic:
@leokondrashov: I have addressed all your comments.

leokondrashov (approved):

PR #531, opened 2024-10-24
Integrating vSwarm with InVitro

KarthikL1729:
## Summary

This PR is towards integrating the InVitro deployer and loader to the vSwarm benchmark suite. This is required as InVitro currently only uses dummy functions for deployments, and the vSwarm benchmark suite models realistic serverless loads which would help improve the versatility of InVitro.

The deployer (Knative), invoker and driver have mainly been changed, and a new tool called the "mapper" has been added. The mapper reads the input trace files and maps the appropriate proxy functions from vSwarm based on the memory usage and function duration. Zip files for the vSwarm function yamls and their profiles have also been zipped and added as they are required to map and deploy the functions. vSwarm has been added as an option in the config files to allow the user to choose between using standard dummy functions from InVitro or the realistic functions from vSwarm.

The invoker has also been changed to include a new greeter client as the vSwarm functions require a SayHello gRPC call to run as opposed to the executeRPC function call. The recording of the successful and failed requests has also been modified to only count the requests after the warmup duration is done.

In summary, this PR adds a new tool (`tools/mapper`), zipped yamls for the vSwarm functions (`workloads/container/vSwarm_yamls.tar.gz`), modifies the `driver` and `cmd/loader.go` files to integrate vSwarm as an option.

An automatic deployer script has also been added to allow for multiple run sweeps.

* N/A

* N/A


leokondrashov (commented):
Adding 12k lines????
Please consider removing them from the PR or explaining why you cannot go without them. It's impossible to review such a big PR: github webUI freezes from time to time.

Please update the PR description to be more verbose on what, how, and why you do the things in the PR, as it is specified in the PR template.

.gitattributes
leokondrashov, line 1: Move gitattributes to respective folders

cmd/loader.go
```diff
@@ -144,7 +144,11 @@ func parseIATDistribution(cfg *config.LoaderConfiguration) (common.IatDistributi
 func parseYAMLSpecification(cfg *config.LoaderConfiguration) string {
 	switch cfg.YAMLSelector {
 	case "container":
-		return "workloads/container/trace_func_go.yaml"
+		if cfg.VSwarm {
```
leokondrashov, line 147: How would loader react if we try to run vSwarm with Firecracker?

cmd/loader.go
```diff
@@ -174,7 +178,7 @@ func runTraceMode(cfg *config.LoaderConfiguration, readIATFromFile bool, justGen
 	yamlPath := parseYAMLSpecification(cfg)
 
 	traceParser := trace.NewAzureParser(cfg.TracePath, durationToParse)
-	functions := traceParser.Parse(cfg.Platform)
+	functions := traceParser.Parse(cfg.Platform)		// Indexed by hashfunction names
```
leokondrashov, line 181: I don't see any indexing here. What does this comment means then?

pkg/common/constants.go
```diff
@@ -83,7 +83,7 @@ const (
 	FailedWarnThreshold = 0.3
 	// FailedTerminateThreshold Terminate experiment if the percentage of failed invocations (e.g., connection timeouts,
 	// function timeouts) is greater than this threshold
-	FailedTerminateThreshold = 0.5
+	FailedTerminateThreshold = 1.1
```
leokondrashov, line 86: Why do you change it here?

pkg/driver/clients/grpc_client.go
```diff
@@ -29,10 +29,12 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"strings"
 	"time"
-
+	"github.com/google/uuid"
```
leokondrashov, line 32: Please run through gofmt before the push

pkg/driver/trace_driver.go
```diff
@@ -74,7 +74,7 @@ func NewDriver(driverConfig *config.Configuration) *Driver {
 // HELPER METHODS
 // ///////////////////////////////////////
 func (d *Driver) outputFilename(name string) string {
-	return fmt.Sprintf("%s_%s_%d.csv", d.Configuration.LoaderConfiguration.OutputPathPrefix, name, d.Configuration.TraceDuration)
+	return fmt.Sprintf("%s_%s_%d_%d.csv", d.Configuration.LoaderConfiguration.OutputPathPrefix, name, d.Configuration.TraceDuration, len(d.Configuration.Functions))
```
leokondrashov, line 77: Why do you need to change the filename format?

leokondrashov (changes requested):
This is a result of my brief look at the code.

Please also update the tests that fail.

leokondrashov:
Please sign off the commits (read contributor's guide in vhive)

cmd/loader.go
KarthikL1729: This was more for my understanding when modifying the code, that the hashfunction is what I would use to address the function that the specific hash function is mapped to. Will remove it.

cmd/loader.go
KarthikL1729: We have not tested with firecracker since I have not modified anything that relates to firecracker specifically. For now, it is only for Knative and so I think adding a condition for vSwarm in the firecracker block to exit and throw an error would be okay?

cmd/loader.go
leokondrashov: You don't use it for indexing. `functions` is a flat array.

pkg/common/constants.go
KarthikL1729: I changed this for testing, but I was not sure why we want the drivers to terminate if the failure rate is over 50%. Shouldn't it be okay to run them even if the rate of failure is above 50%?

pkg/driver/trace_driver.go
KarthikL1729: I changed it because of the automated deployer, which runs a sweep over the number of functions with the same durations. So without this, the same file would be overwritten multiple times.

tools/mapper/trace_parser.py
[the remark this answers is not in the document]
KarthikL1729: The reason this file exists is to convert the csv into Ordered dicts and for having a uniform naming scheme for the columns. We need Ordered dicts for the mapping process, since the memory and duration traces need to be in sync; and the uniform naming scheme is as the trace itself has varying names for the percentiles between durations and memory.

cmd/loader.go
leokondrashov: Yes, that is sensible. Please also update docs which modes are supported, which are not.

tools/mapper/trace_parser.py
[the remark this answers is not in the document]
leokondrashov: Do you know that you can rename the columns in the DataFrame? I don't see your explanation as a valid reason to add another parser to the code.

leokondrashov:
When updating the branch, please use rebase, not the merge option.

pkg/driver/trace_driver.go
leokondrashov: Overall, is the automatic deployer an integral part of this PR? Can the automatic deployer be changed so as not to overwrite the file? I ask this because that messes up the interface with the invitro (which you didn't declare in the PR description BTW) that my analysis code relies on. So, change it only if you actually need it to be different.

pkg/driver/clients/grpc_client.go
```diff
@@ -87,38 +89,63 @@ func (i *grpcInvoker) Invoke(function *common.Function, runtimeSpec *common.Runt
...
+			return false, record
+		}
+		record.ResponseTime = time.Since(start).Microseconds()
+		record.ActualDuration = uint32(record.ResponseTime)
+		record.Instance = extractInstanceName(response.GetMessage())
+		if strings.HasPrefix(response.GetMessage(), "FAILURE - mem_alloc") {
+			record.MemoryAllocationTimeout = true
+		} else {
+			record.ActualMemoryUsage = common.Kib2Mib(0) //common.Kib2Mib(response.MemoryUsageInKb)
+		}
```
leokondrashov, line 146: Extract this code into a new vSwarmGrpcInvoker instead of making a special case for the grpcInvoker

pkg/driver/deployment/knative.go
```diff
@@ -83,11 +87,63 @@ func knativeDeploySingleFunction(function *common.Function, yamlPath string, isP
 		// second, then round to an integer as that is what the knative config expects
 	}
 
+	// Default is the function name, but if the function is a vSwarm function, the proxy name is used
+	proxyName := function.Name
+
+	if vSwarm {
+		// Read and unmarshal the mapper output file into a map
+
+		mapperFile, err := os.ReadFile("tools/mapper/output.json")
```
leokondrashov, line 96: Can you make it without hard-coding the path to the file? For example, it might be more useful to move mapper's output to the trace directory instead of using the global file for the mappings.

cmd/loader.go
```diff
@@ -145,8 +145,15 @@ func parseIATDistribution(cfg *config.LoaderConfiguration) (common.IatDistributi
 func parseYAMLSpecification(cfg *config.LoaderConfiguration) string {
 	switch cfg.YAMLSelector {
 	case "container":
-		return "workloads/container/trace_func_go.yaml"
+		if cfg.VSwarm {
+			return "workloads/container/yamls/"
```
leokondrashov, line 149: Rename this directory to vSwarm yamls

pkg/driver/deployment/knative.go
```diff
@@ -83,11 +87,63 @@ func knativeDeploySingleFunction(function *common.Function, yamlPath string, isP
...
+		proxyName = mapperOutput[function.HashFunction]["proxy-function"]
+		log.Info("Proxy name: ", proxyName)
+		// Read the deployment info file for yaml locations and predeployment commands if any
+
+		deploymentInfoFile, err := os.ReadFile(yamlPath + "deploy_info.json")
+		if err != nil {
+			log.Warn("No deployment info file")
+		}
+
+		var deploymentInfo map[string]map[string]interface{}
```
leokondrashov, line 116: Why do you use this data structure? Can you use sensible structs for that?

pkg/driver/deployment/knative.go
```diff
@@ -83,11 +87,63 @@ func knativeDeploySingleFunction(function *common.Function, yamlPath string, isP
...
+		if err != nil {
+			log.Warn("Failed to unmarshal mapper output file")
+		}
+
+		// Find the proxy name for the function
+		proxyName = mapperOutput[function.HashFunction]["proxy-function"]
+		log.Info("Proxy name: ", proxyName)
+		// Read the deployment info file for yaml locations and predeployment commands if any
+
+		deploymentInfoFile, err := os.ReadFile(yamlPath + "deploy_info.json")
```
leokondrashov, line 111: This file seems important for understanding the code. Can you pull it out of the tar.gz?

tools/mapper/trace_parser.py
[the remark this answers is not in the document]
KarthikL1729: The name change is not the main reason. The main reason is populating an ordered dict, to make sure that the durations and memory traces are preserved in order.

pkg/driver/deployment/knative.go
KarthikL1729: This data structure is to unmarshal the json file that has the predeployment commands that are required for deploying functions that require databases.

leokondrashov: Can you use structure of structures instead of map of maps?

pkg/driver/trace_driver.go
KarthikL1729: Will change it back and try to figure something else out for the automatic deployer.

pkg/driver/deployment/knative.go
KarthikL1729: Should I define a new struct for this? I thought this is a pretty simple interface to unmarshal into and that adding a specific struct for this would complicate things.

leokondrashov:
Will be splitted into parts and reviewed separately

PR #541, opened 2024-10-30
Added the mapper tool to map trace to vSwarm proxies

KarthikL1729:
Added documentation for the mapper tool.

## Summary

The mapper tool is used to map the Azure trace functions to the vSwarm proxy functions based on the profiles of vSwarm functions and their correlation with the Azure trace function durations and memory utilization.

This PR adds the mapper tool, which loads the function duration and memory trace with Pandas, finds the optimal mapping based on modelling the problem as a linear sum assignment, and then outputs a file which maps each HashFunction in the trace to a proxy function in vSwarm. The profile needs to be unzipped before using the mapper, and the profile is included in a tar.gz file in the mapper directory.

* N/A

* N/A


tools/mapper/find_proxy_function.py
```diff
@@ -0,0 +1,252 @@
...
+            trace_functions=trace_functions, proxy_functions=proxy_functions
+        )
+        if err == -1:
+            log.error(
+                f"One-To-One mapping between trace function and proxy function not obtained"
+            )
+            log.info(
+                f"Getting closest proxy function for every trace function. Note: Mapping may not be unique"
+            )
+            trace_functions, err = get_closest_proxy_function(
```
leokondrashov, line 225: That is extremely unexpected behavior. If something goes sideways in the unique assignment, I'll get a non-unique one. I would be pretty annoyed to find that my tool tried to be too smart about the operation that I asked it to do. If operation is not possible, the tool should just exit.

tools/mapper/find_proxy_function.py
```diff
@@ -0,0 +1,252 @@
...
+    
+    Returns:
+    - `dict`: Dictionary containing information regarding trace functions with the associated proxy functions
+    - `int`: 0 if no error. -1 if error
+    """
+
+    try:
+
+        trace_functions = OrderedDict(trace_functions)
+        proxy_functions = OrderedDict(proxy_functions)
```
leokondrashov, line 65: Again, why do you need ordereddict here?

tools/mapper/find_proxy_function.py
```diff
@@ -0,0 +1,252 @@
...
+        # Go through the assignment solution
+        for assignment in assignments:
+            row_index = assignment[0]
+            col_index = assignment[1]
+            trace = ""
+            proxy = ""
+            for tf in trace_functions:
+                if row_index == trace_functions[tf]["index"]:
+                    trace = tf
+                    break
```
leokondrashov, line 101: You linearly search in the dict. Are sure that this data structure is suitable for the task? You also literally created trace_list that has all of them in the index order.

leokondrashov (changes requested):
This a very brief look into a single function of the mapper. Please fix all of possible problems before submitting the PR.

docs/mapper_cdf.png
JooyoungPark73, line 1: Please change the axis label to CDF

docs/mapper.md
```diff
@@ -0,0 +1,134 @@
...
+| Average Memory (MB) | 142.66478734351986 MB |
+
+### Error metrics and mapping accuracy 
+
+We have evaluated the accuracy and effectiveness of the mapper tool using the following metrics:
+
+- Duration CDF plot of the trace functions and the mapping that is generated
+- Wasserstein distance (WD) plot between the trace functions and the mapped functions
+
+As mentioned previously, the maximum function duration that we can achieve with vSwarm functions is near 27 seconds, and so, for the WD plot, we exclude trace functions that are beyond this duration to have a better comparison of the mapping accuracy.
```
JooyoungPark73, line 107: Are you planning to mention that you will fill mock functions for the functions that are not mapped?

JooyoungPark73 (changes requested):

docs/mapper.md
KarthikL1729: I've mentioned that previously in the doc, line 63.

JooyoungPark73 (dismissed):

requirements.txt
```diff
@@ -6,4 +6,6 @@ pytest==8.3.2
 cloudpickle==3.1.0
 seaborn==0.13.0
 tqdm==4.67.0
-statsmodels==0.14.0
\ No newline at end of file
+statsmodels==0.14.0
+argparse
+typing
```
leokondrashov, line 11: Please fix the version to specific ones

docs/mapper.md
```diff
@@ -0,0 +1,135 @@
...
+
+The [`sampler`](https://github.com/vhive-serverless/invitro/tree/main/sampler) tool in InVitro can be used to generate the sampled traces from the original Azure traces.
+
+For every function in the trace, the closest function in the [`vSwarm`](https://github.com/vhive-serverless/vSwarm/tree/main/) benchmark suite is set as its proxy (50-percentile memory and 50-percentile duration are considered to find the highest correlation). The 50th percentile is used to ensure that the mapping is not only corresponding to the peak values of the workload, but is also leading to a representative proxy function. Currently the tool utilizes only _Serving Functions_ that are _NOT Pipelined_ as proxy functions.
+
+Currently, vSwarm does not have full coverage of the Azure trace functions. To make sure that we do not have a high error in the mapping of functions, we set a hard threshold of 40% as the maximum absolute error (from the actual trace function duration) that a proxy function can have to be mapped to a trace function. If we do not find an eligible function for the mapping from vSwarm, we fall back to using standard InVitro sample functions in their place for those functions alone.
```
leokondrashov, line 63:
```suggestion
Currently, vSwarm does not have full coverage of the Azure trace functions. To make sure that we do not have a high error in the mapping of functions, we set a hard threshold of 40% as the maximum absolute error (from the actual trace function duration) that a proxy function can have to be mapped to a trace function. If we do not find an eligible function for the mapping from vSwarm, we fall back to using standard InVitro trace functions in their place for those functions alone.
```

docs/mapper.md
```diff
@@ -0,0 +1,135 @@
...
+- Average memory error
+- Average duration error
+- Average absolute memory error
+- Average absolute duration error
+- Average relative memory error
+- Average relative duration error
+- Average absolute relative memory error
+- Average absolute relative duration error
+- Functions in the trace with 0 duration
+- Number of mapped functions with higher than 40% duration error (replaced by InVitro sample functions)
```
leokondrashov, line 133: Can you provide reference numbers for the full trace mapping? That would be helpful to see the ballpark of the possible errors.

docs/mapper.md
leokondrashov, line 1: Overall, please reduce the amount of passive voice in docs

tools/mapper/mapper.py
```diff
@@ -0,0 +1,286 @@
...
+        log.critical(f"Load Generation failed")
+        return
+
+    # Getting a proxy function for every trace function
+    trace_functions, err = get_proxy_function(
+        trace_functions=trace_functions,
+        proxy_functions=proxy_functions,
+    )
+    if err == -1:
+        log.critical(f"Load Generation failed")
```
leokondrashov, line 194: Please work on the error messages. In most cases, message says "Load Generation failed". I have no idea what that means in the context of mapper, it doesn't generate the load. Overall, you have a ton of error checking which end up with the same error message from user perspective.

tools/mapper/mapper.py
```diff
@@ -0,0 +1,286 @@
...
+                f"Mapper output file for trace functions {output_filepath} cannot be read. Error: {e}"
+            )
+            log.critical(f"Load Generation failed")
+            return
+    else:
+        log.critical(f"Mapper output file for trace functions {output_filepath} not found")
+        log.critical(f"Load Generation failed")
+        return
+    
+    if invoke:
```
leokondrashov, line 107: This parameter has confusing name. We don't invoke from `generate_plot`.

tools/mapper/mapper.py
```diff
@@ -0,0 +1,286 @@
...
+        mapped_durations = []
+        for trace in trace_functions:
+            duration = trace_functions[trace]["duration"]["50-percentile"]
+            if duration > 27000:
+                continue
+            proxy_name = mapped_traces[trace]["proxy-function"]
+            profile_duration = proxy_functions[proxy_name]["duration"]["50-percentile"]
+            trace_durations.append(duration)
+            mapped_durations.append(profile_duration)
+        return trace_durations, mapped_durations
```
leokondrashov, line 139: A lot of duplicated code and repeated hard-coded constant

docs/mapper.md
```diff
@@ -0,0 +1,135 @@
+# Mapper
+
+The mapper tool can be used to map the functions in a given trace directory (with memory and duration traces) to the proxy functions in the [`vSwarm`](https://github.com/vhive-serverless/vSwarm/tree/main/) benchmark suite. The benchmarks present in the vSwarm benchmark suite have been profiled and their memory utilization and duration traces have been collected and stored in the `profile.json` file. Each function in the trace is mapped to a function in the benchmark suite as its closest proxy (based on memory and duration correlation).
+
+The `profile.json` JSON output file is generated by the [`profiler` tool](https://github.com/vhive-serverless/vSwarm/tree/load-generator/tools/profiler#profiler) to obtain the profile of the benchmark suite functions.
+
+### Usage
+
+```bash
+usage: mapper.py [-h] -t TRACE_DIRECTORYPATH -p PROFILE_FILEPATH [-o OUTPUT_FILEPATH] [-u UNIQUE_ASSIGNMENT]
```
leokondrashov, line 10: This line is inconsistent with the rest of the help message.

tools/mapper/find_proxy_function.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+
+        for function_name in proxy_functions:
+            del proxy_functions[function_name]["index"]
+
+        return trace_functions, 0
+
+    except Exception as e:
+        log.error(f"Finding closest proxy function failed. Error: {e}")
+        return trace_functions, -1
```
leokondrashov, line 96: What exceptions are expected here?

tools/mapper/find_proxy_function.py
```diff
@@ -0,0 +1,129 @@
...
+    - `trace_functions` (dict): Dictionary containing information regarding trace functions
+    - `proxy_functions` (dict): Dictionary containing information regarding proxy functions
+    
+    Returns:
+    - `dict`: Dictionary containing information regarding trace functions with the associated proxy functions
+    - `int`: 0 if no error. -1 if error
+    """
+
+    log.info(
+        f"Lower the correlation value, the proxy function is a better proxy of the trace function"
```
leokondrashov, line 115: What does this line mean?

tools/mapper/mapper.py
KarthikL1729: This code is from the WD distance branch, and so I mainly used it for the metric generation. Shall I remove it entirely? Since the mapper does not use it.

JooyoungPark73 (approved):

tools/mapper/test_files/trace.tar.gz
leokondrashov, line 1: This is a very small archive used only for tests. How big are the files inside? If it is comparable with the example trace, we can just add the files directly. It would be even better if you could use the example trace for the test.

.github/workflows/tools-tests.yaml
```diff
@@ -35,3 +35,47 @@ jobs:
...
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ hashFiles('requirements.txt') }}
+
+      - name: Install requirements
+        run: pip install -r ./requirements.txt
+
+      - name: Profile load check
+        run: |
+          tar -xzvf tools/mapper/profile.tar.gz -C tools/mapper/
+          python3 -c "import json; json.load(open('tools/mapper/profile.json'))"
```
leokondrashov, line 68: Wait, is the profile.json the only file in the archive? If so, can you use `gzip` directly? `tar` is commonly used to archive folders.

.github/workflows/tools-tests.yaml
```diff
@@ -35,3 +35,47 @@ jobs:
...
+          tar -xzvf tools/mapper/test_files/trace.tar.gz -C tools/mapper/test_files/
+          python3 tools/mapper/trace_load_test.py -t tools/mapper/test_files/extremes/
+
+      - name: Extreme mapping tests
+        run: |
+          python3 tools/mapper/mapper.py -t tools/mapper/test_files/extremes -p tools/mapper/profile.json
+
+      - name: Run mapper tool on sample trace
+        run: |
+          python3 tools/mapper/mapper.py -t tools/mapper/test_files/20 -p tools/mapper/profile.json
```
JooyoungPark73, line 81: Please compare with sample correct mapping

JooyoungPark73 (changes requested):

tools/mapper/mapper.py
```diff
@@ -0,0 +1,244 @@
...
+    log.info(f"Average absolute relative duration error: {abs_rel_dur_error/total_functions}")
+    log.info(f"Duration errors: {dur_count}")
+    log.info(f"Functions with 0 duration: {zero_duration}")
+
+    log.info(f"Replacing the functions with high duration error with invitro functions.")
+
+    ## Write the updated mapper output
+
+    try:
+        with open(output_filepath, "w") as jf:
```
leokondrashov, line 235: Why do you write it twice? I think it is better to add a stage of filtering and error counting on the same data structure (better to extract to separate functions with tests, preferably).

tools/mapper/mapper.py
```diff
@@ -0,0 +1,244 @@
...
+    trace_functions = {}
+
+    for _, row in duration_info.iterrows():
+        hash_function = row["HashFunction"]
+        trace_functions[hash_function] = {}
+        trace_functions[hash_function]["duration"] = row
+
+    for _, row in memory_info.iterrows():
+        hash_function = row["HashFunction"]
+        trace_functions[hash_function]["memory"] = row
```
leokondrashov, line 64: Why do you need to create the dict from dataframe? All these actions can be done with `.set_index` and `.concat`

tools/mapper/mapper.py
leokondrashov: If you have a thoughts that it is useful, leave and document it. If it is completely useless, remove

tools/mapper/mapper.py
KarthikL1729: I wanted to return a unified data structure from the load trace with the duration data and memory data clearly separated. Before this step the duration and memory data re in two separate dataframes, so I would still need to make a new consolidated dataframe, and I thought a dict would be simpler. Would the consolidated dataframe be preferred?

leokondrashov: You can join dataframes by index (BTW, hash function is not guaranteed to be unique. You should use (owner, app, function) to be sure). You can add suffixes for the columns with the same name so you can address them individually.

docs/mapper.md
```diff
@@ -0,0 +1,138 @@
+# Mapper
+
+Use the mapper tool to map functions in a specific trace directory (with memory and duration traces) to proxy functions in the [`vSwarm`](https://github.com/vhive-serverless/vSwarm/tree/main/) benchmark suite. The benchmarks in the vSwarm suite have been profiled, and their memory utilization and duration traces are stored in the `profile.json` file. The tool maps each function in the trace to its closest proxy in the benchmark suite based on memory and duration correlation.
+
+The [`profiler` tool](https://github.com/vhive-serverless/vSwarm/tree/load-generator/tools/profiler#profiler) generates the `profile.json` JSON output file to profile the benchmark suite functions.
+
+### Usage
```
leokondrashov, line 7: Please put the note about unzipping the profile.json

leokondrashov (approved):
Looks fine to me. @cvetkovic Please take a look

JooyoungPark73 (approved):

docs/mapper_cdf.png
cvetkovic, line 1: What do the tail and sharp drop from 10^4 to 2.5*10^6 mean?

KarthikL1729: That is because the vSwarm functions have a maximum of 27 seconds run time. The current graph shows the mapping accuracy up to 10^6 seconds, but drops the functions that are longer than 27 seconds in the sampled trace. I'll update that to not drop the functions beyond 27 seconds, and then the graph should be clearer.

KarthikL1729: I am unsure about what the sharp drop is, I think that is an artifact of using plt.hist from matplotlib.

leokondrashov: You can cut it with xlim, to leave the drop is outside of the plot, so no one is confused about that

leokondrashov:
Looks good to me. Please squash the commits and rebase on main branch properly. @cvetkovic approve it if you are good with that.

cvetkovic (dismissed):

leokondrashov (approved):

PR #557, opened 2024-11-15
RPS mode with cold/warm ratio

cvetkovic:
- Support for RPS mode
- Reimplementation of individual trace driver
- Specification generator bugfixing
- Code stability improvement (a lot of new unit and integration tests)
- Workload code deduplication
- Knative deployment parallelization
- Code moving and beautification


cvetkovic:
@leokondrashov: This is the last PR from the big one. Please have a look thoroughly at this once, since it affects the core functionality of the loader. After this one, I will follow up with a new one that will extract Dirigent-specific features/settings.

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,376 @@
...
+				{13_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
+				{14_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
+				{15_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
+				{16_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
+				{17_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
+				{18_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
+				{19_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000, 10_000_000},
```
leokondrashov, line 137: Doesn't see right. Sum of all of them is 129s, which is bigger than expected 2min.

pkg/common/specification_types.go
```diff
@@ -24,8 +24,7 @@
 
 package common
 
-// IATMatrix - columns are minutes, rows are IATs
-type IATMatrix [][]float64
+type IATArray []float64
```
leokondrashov, line 27: I think we should create an explanation how exactly it is interpreted. For example, should we expect invocation before the first IAT, how would it correspond to the PerMinuteCounts. I think some inconsistencies in the tests and code come from that misunderstanding between how it was before (with shift option for example) and now.

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,376 @@
...
+		expectedIAT        common.IATArray
+		expectedCount      []int
+	}{
+		{
+			testName:           "2min_1rps",
+			experimentDuration: 2,
+			rpsTarget:          1,
+			expectedIAT: []float64{
+				// minute 1
+				1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000, 1_000_000,
```
leokondrashov, line 23: I would expect this test to lead to the same output as in trace mode with per-minute counts of `60, 60` with equidistant IAT, but in the trace tests, we expect the IATs to start with 0. One of them needs to be changed.

pkg/generator/specification.go
```diff
@@ -118,51 +126,50 @@ func (s *SpecificationGenerator) generateIATPerGranularity(numberOfInvocations i
...
+	shiftIAT bool, granularity common.TraceGranularity) (common.IATArray, []int, common.ProbabilisticDuration) {
+
+	var IAT []float64
+	var perMinuteCount []int
 	var nonScaledDuration []float64
 
 	numberOfMinutes := len(invocationsPerMinute)
 	for i := 0; i < numberOfMinutes; i++ {
-		minuteIAT, duration := s.generateIATPerGranularity(invocationsPerMinute[i], iatDistribution, shiftIAT, granularity)
+		minuteIAT, duration := s.generateIATPerGranularity(i, invocationsPerMinute[i], iatDistribution, shiftIAT, granularity)
```
leokondrashov, line 144: When we have 0 invocations per minute, we return the array with a single 0. It doesn't seem right to append this to IAT array. There should be IAT > 60s, instead.

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,376 @@
...
+		rpsTarget          float64
+		cooldownSeconds    int
+		expectedIAT        []common.IATArray
+		expectedCount      [][]int
+	}{
+		{
+			testName:           "2min_1rps",
+			experimentDuration: 2,
+			rpsTarget:          1,
+			cooldownSeconds:    10,
```
leokondrashov, line 126: We should add the test for the cooldown period of > 60s. Same for RPS target of <1/60 RPS.

leokondrashov (commented):
We should clarify the interpretation of the IAT array; only then can I check the driver and spec generation code. Right now, it seems inconsistent in the tests for RPS and trace modes.

pkg/driver/deployment/knative.go
```diff
@@ -46,15 +48,25 @@ func newKnativeDeployerConfiguration(cfg *config.Configuration) knativeDeploymen
...
 	knativeConfig := newKnativeDeployerConfiguration(cfg)
 
+	queue := make(chan struct{}, runtime.NumCPU()) // message queue as a sync method
+	deployed := sync.WaitGroup{}
+	deployed.Add(len(cfg.Functions))
+
 	for i := 0; i < len(cfg.Functions); i++ {
+		queue <- struct{}{}
+
 		knativeDeploySingleFunction(
```
leokondrashov, line 58: You missed the `go` in front of this call. The point is to do that in goroutines.

pkg/trace/parser.go
```diff
@@ -42,14 +43,16 @@ import (
...
 type AzureTraceParser struct {
 	DirectoryPath string
+	YAMLPath      string
 
 	duration              int
 	functionNameGenerator *rand.Rand
 }
 
-func NewAzureParser(directoryPath string, totalDuration int) *AzureTraceParser {
+func NewAzureParser(directoryPath string, yamlPath string, totalDuration int) *AzureTraceParser {
```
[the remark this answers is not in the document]
cvetkovic: Done in https://github.com/vhive-serverless/invitro/pull/557/commits/6a65a82e35a4f4f116ac0bceb5a9f0f297dcf077.

pkg/driver/trace_driver.go
```diff
@@ -634,34 +508,75 @@ func (d *Driver) internalRun(iatOnly bool, generated bool) {
...
 	}
 
+	if readIATsFromFile {
+		for i := range d.Configuration.Functions {
+			var spec common.FunctionSpecification
+
+			iatFile, _ := os.ReadFile("iat" + strconv.Itoa(i) + ".json")
+			err := json.Unmarshal(iatFile, &spec)
+			if err != nil {
+				log.Fatalf("Failed tu unmarshal iat file: %s", err)
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/cb757dae4d84cf5e7a813660d00627e3db6eb7ed.

pkg/driver/trace_driver.go
```diff
@@ -634,34 +508,75 @@ func (d *Driver) internalRun(iatOnly bool, generated bool) {
...
+	}
+}
+
+func (d *Driver) RunExperiment(generateSpecs bool, writeIATsToFile bool, readIATsFromFile bool) {
+	if generateSpecs && readIATsFromFile {
+		log.Fatal("Invalid loader configuration. Cannot be forced to generate IATs and read the from file in the same experiment.")
+	}
+
+	if generateSpecs {
+		d.generateSpecs()
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/95876587e79ea0bcec1e240c33f74bf23d3c0a00.

pkg/driver/trace_driver.go
```diff
@@ -634,34 +508,75 @@ func (d *Driver) internalRun(iatOnly bool, generated bool) {
...
+	for i, function := range d.Configuration.Functions {
+		file, _ := json.MarshalIndent(function.Specification, "", " ")
+		err := os.WriteFile("iat"+strconv.Itoa(i)+".json", file, 0644)
+		if err != nil {
+			log.Fatalf("Writing the loader config file failed: %s", err)
 		}
+	}
+}
+
+func (d *Driver) RunExperiment(generateSpecs bool, writeIATsToFile bool, readIATsFromFile bool) {
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/95876587e79ea0bcec1e240c33f74bf23d3c0a00.

pkg/driver/deployment/knative.go
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/c54df798ee3910fcd72c2a4d6df34fa55a834e7c.

@leokondrashov: Can you double-check that the code is correct now?

pkg/driver/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+				writerDone.Wait()
+				allRecordsWritten.Done()
+
+				return
+			}
+		}
+	}
+}
+
+func (d *Driver) createGlobalMetricsCollector(filename string, collector chan *mc.ExecutionRecord,
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/8558bb7e70752a5d8ad671bd6fed9bdcbbe6df0b.

pkg/driver/trace_driver.go
```diff
@@ -263,11 +200,13 @@ func (d *Driver) functionsDriver(list *list.List, announceFunctionDone *sync.Wai
 	var currentPhase = common.ExecutionPhase
 
 	waitForInvocations := sync.WaitGroup{}
+	currentMinute, currentSum := 0, 0
```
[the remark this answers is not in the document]
cvetkovic: Renamed to `invokedSinceExperimentStarted`.

pkg/generator/specification.go
```diff
@@ -48,15 +48,23 @@ func NewSpecificationGenerator(seed int64) *SpecificationGenerator {
...
+	endIndex := numberOfInvocations
 	totalDuration := 0.0 // total non-scaled duration
 
-	for i := 0; i < numberOfInvocations; i++ {
+	if minuteIndex == 0 {
+		iatResult = []float64{0.0}
+		endIndex = numberOfInvocations - 1
+	}
+
+	// -1 because the first invocation happens at the beginning of minute
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/488743533be8cb0c567859a90c73145a1afbabae.

pkg/generator/specification_test.go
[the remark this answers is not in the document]
cvetkovic: Done in https://github.com/vhive-serverless/invitro/pull/557/commits/6d29c101488f90ffdb08adcbda36cb7d06987ca7.

pkg/generator/specification_test.go
```diff
@@ -114,16 +114,16 @@ func TestSerialGenerateIAT(t *testing.T) {
...
-			expectedPoints:   [][]float64{{0, 60000000}},
+			expectedPoints:   []float64{0},
 			testDistribution: false,
 		},
 		{
 			testName:         "one_invocations_exponential_shift",
 			invocations:      []int{1},
 			iatDistribution:  common.Exponential,
-			shiftIAT:         true,
+			shiftIAT:         false,
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/6d29c101488f90ffdb08adcbda36cb7d06987ca7.

pkg/driver/trace_driver_test.go
```diff
@@ -407,21 +409,17 @@ func TestDriverCompletely(t *testing.T) {
...
 			}
 
-			expectedInvocations := 5
-			if test.withWarmup {
-				expectedInvocations = 10
-			}
-
-			if !(successfulInvocation == expectedInvocations && failedInvocations == 0) {
+			expectedInvocations := test.expectedInvocations
+			if !(successfulInvocation >= expectedInvocations && failedInvocations == 0) {
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/6d29c101488f90ffdb08adcbda36cb7d06987ca7.

pkg/driver/trace_driver_test.go
```diff
@@ -337,33 +333,39 @@ func TestDriverBackgroundProcesses(t *testing.T) {
...
+			testName:            "without_warmup",
+			withWarmup:          false,
+			expectedInvocations: 5,
 		},
 		{
-			testName:   "with_warmup",
-			withWarmup: true,
+			testName:            "with_warmup",
+			withWarmup:          true,
+			expectedInvocations: 9,
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/6d29c101488f90ffdb08adcbda36cb7d06987ca7.

pkg/generator/specification.go
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/6d29c101488f90ffdb08adcbda36cb7d06987ca7.

pkg/driver/trace_driver_test.go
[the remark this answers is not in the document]
cvetkovic: There was an additional problem with the loader. See https://github.com/vhive-serverless/invitro/pull/557/commits/050d5ac20f7fa9f6478aa4be40d01b1965d63f25.

pkg/generator/rps_test.go
cvetkovic: RPS mode works slightly differently. Concretely, this is a 100% cold start experiment with RPS=1. Because Knative's minimum autoscaling stable window is 6s, the minimum keep-alive for a function is 6s. This means that we need multiple functions to achieve RPS=1, each scaling up/and down with a 1-second delay. In RPS mode, the number of functions is determined by the `RpsCooldownSeconds` parameter. This produces the IAT array as above. Due to the math complexity, the cold start experiment sleeps for the first ten seconds. In the results, the user should discard the first and the last `RpsCooldownSeconds` of the results, since the RPS at those points is lower than the requested one.

I will add the paragraph above to the documentation and to the comments in the code.

pkg/generator/rps_test.go
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/07875f8d12ab1b8aa750317c6c0b0a02f55a5d34.

pkg/generator/rps_test.go
cvetkovic: I don't think this makes sense for less than 6 seconds since the minimum stable window period in Knative is 6s. 

This parameter is just there to determine the number of functions needed to deploy to generate a cold start trace of requested RPS, primarily because of Knative control plane issues. A higher value of this parameter is needed for higher cold start RPS values in Knative.

pkg/workload/openwhisk/workload_openwhisk.go
```diff
@@ -24,13 +24,74 @@
...
+		result["body"] = obj
+		return result
+	}
+
+	ts, _ := strconv.Atoi(requestedCpu)
+
+	start := time.Now()
+	timeLeftMilliseconds := uint32(ts)
+
+	timeConsumedMilliseconds := uint32(time.Since(start).Milliseconds())
```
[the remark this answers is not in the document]
cvetkovic: Addressed in https://github.com/vhive-serverless/invitro/pull/557/commits/389e81fe310793463a3735076a133b8e985796c8.

pkg/generator/specification_test.go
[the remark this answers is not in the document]
cvetkovic: Addressed in https://github.com/vhive-serverless/invitro/pull/557/commits/f8f2f132a17d4f273164d6b81a73f1dacc67b39e. Please have a look.

pkg/generator/specification.go
```diff
@@ -113,56 +125,78 @@ func (s *SpecificationGenerator) generateIATPerGranularity(numberOfInvocations i
...
+			continue
+		} else if accumulatedIdle != 0 {
+			IAT = append(IAT, accumulatedIdle)
+			IAT = append(IAT, minuteIAT[1:]...)
+			perMinuteCount = append(perMinuteCount, len(minuteIAT)-1)
+			accumulatedIdle = 0.0
+		} else {
+			IAT = append(IAT, minuteIAT...)
+			perMinuteCount = append(perMinuteCount, len(minuteIAT))
+		}
```
leokondrashov, line 173: I think we can do much simpler with this (and restore the `generateIATPerGranularity` to the old version):
```suggestion
		minuteIAT, duration := s.generateIATPerGranularity(invocationsPerMinute[i], iatDistribution, shiftIAT, granularity)
		IAT[len(IAT)-1] += minuteIAT[0]
		IAT = append(IAT, minuteIAT[1:]...)
```

If we add `return []float64{getBlankTimeUnit}, 0.0` in case of no invocation in the period and start with `IAT:=[]float64{0.0}`.

On the minute border, real IAT is the last element of the previous minute + the first element of the next one. That's because the last element in the array is the interval between the last invocation and the end of the minute, while the first one is between the start of the minute and the first invocation. So, the method would report the interval between the arrivals. With the empty minutes, the last element of the IAT array would play the role of `accumulateIdle`, summing all the empty minutes between invocations.

leokondrashov (commented):
I looked over test cases, I don't think they are right, the expected values don't make sense. I've proposed the solution, it should work as I outlined in the description.

pkg/generator/specification_test.go
```diff
@@ -247,120 +211,123 @@ func TestSerialGenerateIAT(t *testing.T) {
...
+				12_000_000,
+				// minute 2
+				15_000_000,
+				15_000_000,
+				15_000_000,
+				15_000_000,
+				// minute 5
+				120_000_000,
+				// minute 9
+				180_000_000,
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/dff61bc6d688d41c0277dc703c54a9aaab778979.

pkg/generator/specification_test.go
```diff
@@ -247,120 +211,123 @@ func TestSerialGenerateIAT(t *testing.T) {
...
+			iatDistribution: common.Equidistant,
+			shiftIAT:        false,
+			granularity:     common.MinuteGranularity,
+			expectedPoints: []float64{
+				60_000_000,
+				12_000_000,
+				12_000_000,
+				12_000_000,
+				12_000_000,
+				60_000_000,
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/dff61bc6d688d41c0277dc703c54a9aaab778979.

pkg/generator/specification_test.go
```diff
@@ -247,120 +211,123 @@ func TestSerialGenerateIAT(t *testing.T) {
...
+				12_000_000,
+				12_000_000,
+				12_000_000,
+			},
+			testDistribution: false,
 		},
 		{
-			testName:        "2sec_5qps_equidistant",
-			invocations:     []int{5, 4, 2},
+			testName:        "2min_5ipm_with_zero_equidistant",
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/dff61bc6d688d41c0277dc703c54a9aaab778979.

pkg/generator/specification_test.go
```diff
@@ -247,120 +211,123 @@ func TestSerialGenerateIAT(t *testing.T) {
...
-				},
-				{
-					// second 3
-					0,
-					500000,
-					500000,
-				},
+			expectedPoints: []float64{
+				1_000_000,
+				1_000_000,
```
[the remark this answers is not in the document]
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/dff61bc6d688d41c0277dc703c54a9aaab778979.

pkg/generator/rps_test.go
cvetkovic: Ohhh. I understand. See changes https://github.com/vhive-serverless/invitro/pull/557/commits/c15a64330d53903695c6b6439d8c767fb735043d.

I added a warm start test with RPS=0.01 (`6min_0.01rps`) -- invoke a function every 100s, which is even lower RPS than what you suggested (1/60). I picked 100s because of math rounding.

For cold stars, I added functions:
- with 10s cooldown and 0.01 RPS (`6min_0.01rps_10s_cooldown`). 
- with 120s cooldown and 0.0.1 RPS (`6min_0.01rps_120s_cooldown`) - loader figures out it should create two functions here.

pkg/generator/specification.go
```diff
@@ -50,13 +50,18 @@ func NewSpecificationGenerator(seed int64) *SpecificationGenerator {
...
 		return []float64{}, 0.0
 	}
 
 	var iatResult []float64
 	totalDuration := 0.0 // total non-scaled duration
 
-	for i := 0; i < numberOfInvocations; i++ {
+	// first invocation at the beginning of minute
+	iatResult = []float64{0.0}
+	endIndex := numberOfInvocations - 1
```
leokondrashov, line 62: Because of that, non-equidistant would break: with two invocations, it would always return [0, 60], while the invocation should be somewhere in the middle.

pkg/generator/specification.go
```diff
@@ -113,56 +122,100 @@ func (s *SpecificationGenerator) generateIATPerGranularity(numberOfInvocations i
...
+			continue
+		} else {
+			// if perMinuteCount for last non-empty minute == 1
+			// then X = getBlankTimeUnit(granularity)	=> case when minute was not split
+			// else X = IAT[len(IAT) - 1]				=> case when minute was split
+			// minuteIAT[0] = emptyTime + X
+
+			minuteIAT[0] += emptyTime
+			if !firstInvocation && perMinuteCount[lastNonZeroIndex] != 1 {
+				minuteIAT[0] += IAT[len(IAT)-1]
```
leokondrashov, line 178: This doesn't seem right. This works exclusively for the equidistant now, it won't work for any other IAT.

You changed the generateIATPerGranularity in a way that it doesn't add up to a full minute (you can see that in tests that you added). So, the last value in minuteIAT doesn't correspond to the interval after the last invocation to the end of minute, but interval between two last invocations. In case of equidistant, they are the same, so the tests work fine. But there are no tests for other IATs with several minutes of invocations.

I propose you return the last element of the array (essentially, just revert the changes to generateIATPerGranularity), and use `IAT[len(IAT)-1]` as the `emptyTime`, as I outlined in previous suggestion yesterday.

pkg/generator/specification.go
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/134b82c6b3a21b974bf84c914b2668cfe2968ff8.

pkg/generator/specification.go
cvetkovic: Fixed in https://github.com/vhive-serverless/invitro/pull/557/commits/134b82c6b3a21b974bf84c914b2668cfe2968ff8.

pkg/generator/specification_test.go
```diff
@@ -64,6 +65,73 @@ var testFunction = common.Function{
...
+		epsilon := 10e-3
+
+		t.Run(testName, func(t *testing.T) {
+			sg := NewSpecificationGenerator(123)
+			data, _ := sg.generateIATPerGranularity(test.count, test.iatDistribution, false, test.granularity)
+
+			failed := false
+			if test.expectedPoints != nil {
+				for i := 0; i < len(test.expectedPoints); i++ {
+					if len(test.expectedPoints) != len(data) {
```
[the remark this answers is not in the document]
leokondrashov: This does a bad job at detecting the wrong number of points when we have 0 expected points. It just skips this check. We also have a wrong test for 0 invocations: we set the expected points to an empty array, while it should be array containing the time granularity (see special case, L54, `specification.go`)

pkg/generator/specification_test.go
```diff
@@ -114,126 +182,96 @@ func TestSerialGenerateIAT(t *testing.T) {
...
+				11_689_078.788397402,
+			},
 			testDistribution: false,
 		},
 		{
-			testName:        "1min_5ipm_equidistant",
-			invocations:     []int{5},
-			iatDistribution: common.Equidistant,
+			testName:        "one_invocations_exponential_shift_2",
+			invocations:     []int{3},
```
leokondrashov, line 201: Name is wrong

cvetkovic: Fixed.

pkg/generator/specification_test.go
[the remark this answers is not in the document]
cvetkovic: Fixed both in https://github.com/vhive-serverless/invitro/pull/557/commits/0843abec995e1e62cb32e7e5826c81d0a57f1aad.

pkg/driver/trace_driver.go
```diff
@@ -634,34 +491,73 @@ func (d *Driver) internalRun(iatOnly bool, generated bool) {
...
+	for i, function := range d.Configuration.Functions {
+		file, _ := json.MarshalIndent(function.Specification, "", " ")
+		err := os.WriteFile("iat"+strconv.Itoa(i)+".json", file, 0644)
+		if err != nil {
+			log.Fatalf("Writing the loader config file failed: %s", err)
+		}
+	}
+}
+
+func (d *Driver) DumpSpecification(writeIATsToFile bool, readIATsFromFile bool) {
```
leokondrashov, line 533: This function does unexpected thing: it acts as an input as well. Base on the name, I would expect it to be only output.

docs/configuration.md
```diff
@@ -55,6 +55,13 @@ Lambda; https://aws.amazon.com/about-aws/whats-new/2018/10/aws-lambda-supports-f
 
 [^6]: Dirigent specific
 
+[^7] Because Knative's minimum autoscaling stable window is 6s, the minimum keep-alive for a function is 6s. This means
+that we need multiple functions to achieve RPS=1, each scaling up/and down with a 1-second delay from each other. In RPS
+mode, the number of functions for the cold start experiment is determined by the `RpsCooldownSeconds` parameter, which
+is the minimum keep-alive. Due to the implementation complexity, the cold start experiment sleeps for the first
```
leokondrashov, line 61: Same as in documentation in the comment before the functions.

pkg/driver/trace_driver.go
```diff
@@ -276,10 +198,18 @@ func (d *Driver) functionsDriver(list *list.List, announceFunctionDone *sync.Wai
...
+		}
+
+		iatIndex := invokedSinceExperimentStarted + invocationIndex
+
+		if minuteIndex >= totalTraceDuration || iatIndex >= len(IAT) {
 			// Check whether the end of trace has been reached
 			break
-		} else if function.InvocationStats.Invocations[minuteIndex] == 0 {
+		} else if function.Specification.PerMinuteCount[minuteIndex] == 0 {
 			// Sleep for a minute if there are no invocations
```
leokondrashov, line 213: I think that is not needed now, we cover empty minutes by having IATs>60s. I don't think we need to sleep in addition to that.

leokondrashov: I think we should add a test for the empty minute case. In this test, we might check the interval between invocation with the intended interval for several cases in test mode. This would act as proof that we don't drift too much from the generated IATs.

pkg/driver/trace_driver.go
```diff
@@ -307,46 +237,46 @@ func (d *Driver) functionsDriver(list *list.List, announceFunctionDone *sync.Wai
...
+				MinuteIndex:           minuteIndex,
+				InvocationIndex:       invocationIndex,
+				SuccessCount:          &successfulInvocations,
+				FailedCount:           &failedInvocations,
+				FailedCountByMinute:   failedInvocationByMinute,
+				RecordOutputChannel:   recordOutputChannel,
+				AnnounceDoneWG:        &waitForInvocations,
+				AnnounceDoneExe:       addInvocationsToGroup,
+				ReadOpenWhiskMetadata: readOpenWhiskMetadata,
+			}, iatIndex)
```
leokondrashov, line 255: We should push this inside the structure. Also we need to prune the structure: we are passing minute index and invocation index just to create the name (we can do it here and pass it instead); readOpenWhiskMetadata is not used inside now.

pkg/driver/trace_driver.go
```diff
@@ -307,46 +237,46 @@ func (d *Driver) functionsDriver(list *list.List, announceFunctionDone *sync.Wai
...
+				},
+			}
+
+			successfulInvocations++
+		}
+		numberOfIssuedInvocations++
+		invocationIndex++
+
 		if function.InvocationStats.Invocations[minuteIndex] == invocationIndex || hasMinuteExpired(startOfMinute) {
 			readyToBreak := d.proceedToNextMinute(function, &minuteIndex, &invocationIndex, &startOfMinute,
```
leokondrashov, line 274: With current IAT structure, we might not call the proceedToNextMinute on the minute border (e.g., when the IAT=8s in RPS mode). The logic there should be updated to accommodate that.

I noticed that previousIATSum is updated there. Since now we have a flat IAT structure, we might have a problem that this IAT sum doesn't correspond to the sum of elements of IAT array up to i-th one.

leokondrashov (commented):
Some minor issues in the tests for the generator; otherwise, looks good. But I have concerns about the driver now. Since we are not necessarily hit the minute border, we should be careful how it is handled (especially, startOfMinute, previousIATSum, and minuteIndex for empty minutes)

pkg/generator/specification_test.go
```diff
@@ -371,35 +582,32 @@ func TestSerialGenerateIAT(t *testing.T) {
...
-							break
-						}
-						if math.Abs(IAT[min][i]-test.expectedPoints[min][i]) > epsilon {
-							log.Debug(fmt.Sprintf("got: %f, expected: %f\n", IAT[min][i], test.expectedPoints[min][i]))
-
-							failed = true
-							// no break statement for debugging purpose
-						}
+				for i := 0; i < len(test.expectedPoints); i++ {
+					if len(test.expectedPoints) != len(IAT) {
```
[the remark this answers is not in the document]
cvetkovic: I see. I changed it in the `TestGenerateDistribution`, but not in the `TestSerialGenerateIAT`. Fixed now.

pkg/generator/specification_test.go
```diff
@@ -247,120 +285,293 @@ func TestSerialGenerateIAT(t *testing.T) {
...
+				12_000_000, // 5min 12s
+				12_000_000, // 5min 24s
+				12_000_000, // 5min 36s
+				12_000_000, // 5min 48s
+			},
+			testDistribution: false,
+		},
+		{
+			testName:        "2sec_5ipm_with_zero_equidistant_1",
+			invocations:     []int{0, 1, 0, 1},
```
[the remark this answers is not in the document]
cvetkovic: Fixed.

pkg/driver/deployment/knative.go
```diff
@@ -46,15 +48,27 @@ func newKnativeDeployerConfiguration(cfg *config.Configuration) knativeDeploymen
...
 	for i := 0; i < len(cfg.Functions); i++ {
-		knativeDeploySingleFunction(
-			cfg.Functions[i],
-			knativeConfig.YamlPath,
-			knativeConfig.IsPartiallyPanic,
-			knativeConfig.EndpointPort,
-			knativeConfig.AutoscalingMetric,
-		)
+		go func() {
+			queue <- struct{}{}
```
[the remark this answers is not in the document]
cvetkovic: What do you mean by pulling the deployed? What code changes do you suggest?

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,417 @@
...
+		rpsTarget              float64
+		expectedIAT            common.IATArray
+		expectedPerMinuteCount []int
+	}{
+		{
+			testName:               "2min_0rps",
+			experimentDuration:     2,
+			rpsTarget:              0,
+			expectedIAT:            []float64{},
+			expectedPerMinuteCount: []int{},
```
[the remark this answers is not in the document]
leokondrashov: IAT should be empty, but counts should have a number for each minute.

pkg/driver/trace_driver.go
cvetkovic: Fixed.

pkg/driver/deployment/knative.go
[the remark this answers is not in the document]
leokondrashov: Move lines with defer from the knativeDeploySingleFunction; they are right in the beginning. Now, we can put this into the anonymous function used in the goroutine. Essentially, what I'm suggesting is to move the synchronization out of the deployment function.

pkg/driver/trace_driver.go
cvetkovic: Done moving and pruning.

pkg/generator/rps.go
```diff
@@ -0,0 +1,177 @@
...
+	iat := generateFunctionByRPS(experimentDuration, rpsTarget)
+	count := countNumberOfInvocationsPerMinute(experimentDuration, iat)
+	return iat, count
+}
+
+// GenerateColdStartFunctions Because Knative's minimum autoscaling stable window is 6s, the minimum keep-alive for a
+// function is 6s. This means that we need multiple functions to achieve RPS=1, each scaling up/and down with a 1-second
+// delay from each other. In RPS mode, the number of functions for the cold start experiment is determined by the
+// RpsCooldownSeconds parameter, which is the minimum keep-alive. This produces the IAT array as above. Due to the
+// implementation complexity, the cold start experiment sleeps for the first RpsCooldownSeconds seconds. In the results,
```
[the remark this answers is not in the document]
cvetkovic: One should discard results for `max(offset) for all functions` seconds. Have a look at this example. Only after 10s the requested RPS is achieved.
image

I will make this margin time be printed out in the logs and update the documentation accordingly.

pkg/driver/deployment/knative.go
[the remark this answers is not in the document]
cvetkovic: I see. Makes sense. That way we remove synchronization-related arguments from the Knative deployment function.

pkg/generator/rps_test.go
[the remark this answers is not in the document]
cvetkovic: Fixed. Good catch.

pkg/generator/rps.go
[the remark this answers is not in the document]
cvetkovic: Fixed.

docs/configuration.md
cvetkovic: Fixed.

pkg/driver/trace_driver.go
cvetkovic: Have a look at the comment above.

pkg/driver/trace_driver.go
cvetkovic: I think the individual function driver should be rewritten from scratch. Have a look at the last commit where I did that. What do you think about clock drift here? Previously, we were doing adjustments on a minute-basis, now we do it on a per-invocation basis.

pkg/driver/trace_driver.go
```diff
@@ -236,177 +154,136 @@ func (d *Driver) invokeFunction(metadata *InvocationMetadata) {
...
+}
+
+func (d *Driver) functionsDriver(list *list.List, announceFunctionDone *sync.WaitGroup, addInvocationsToGroup *sync.WaitGroup, totalSuccessful *int64, totalFailed *int64, totalIssued *int64, recordOutputChannel chan *mc.ExecutionRecord) {
+	defer announceFunctionDone.Done()
 
 	function := list.Front().Value.(*common.Function)
-	numberOfInvocations := 0
-	for i := 0; i < len(function.InvocationStats.Invocations); i++ {
-		numberOfInvocations += function.InvocationStats.Invocations[i]
+	invocationCount := getTotalInvocationCount(function.Specification.PerMinuteCount, d.Configuration.TraceDuration, d.Configuration.TraceGranularity)
```
leokondrashov, line 181: This number should already be equal to the size of the IAT array by its construction. The only thing that you change is the interpretation of trace duration with second granularity (trace duration is always in minutes). This actually creates a problem with the front end: we parse only ExperimentDuration (+ WarmupDuration + 1) numbers from the trace without multiplying it by 60.

I think part of the confusion is the perMinuteCounts. With second granularity, it is actually per-second ones (we just append it after each granularity, so it is exactly the same as InvocationCounts). But in the getTotalInvocationCount, you interpret them only as per-minute.

pkg/driver/trace_driver.go
```diff
@@ -236,177 +154,136 @@ func (d *Driver) invokeFunction(metadata *InvocationMetadata) {
...
-		currentTime := time.Now()
-		schedulingDelay := currentTime.Sub(startOfMinute).Microseconds() - previousIATSum
-		sleepFor := iat.Microseconds() - schedulingDelay
-		time.Sleep(time.Duration(sleepFor) * time.Microsecond)
+		schedulingDelay := time.Since(startOfIteration)
+		sleepFor := iat - schedulingDelay
+		time.Sleep(sleepFor)
 
-		previousIATSum += iat.Microseconds()
+		startOfIteration = time.Now()
```
leokondrashov, line 223: This would be prone to clock drift over time. I think it is better to have a sum of IAT values from the beginning of the experiment and compare it to the real elapsed time since the beginning of the experiment and correct for the difference. This would correct for the Sleep oversleeping on previous iteration.

pkg/driver/trace_driver.go
```diff
@@ -236,177 +154,136 @@ func (d *Driver) invokeFunction(metadata *InvocationMetadata) {
...
+					InvocationID: invocationID,
+					StartTime:    time.Now().UnixNano(),
+				},
+			}
+
+			successfulInvocations++
+		}
+
+		numberOfIssuedInvocations++
+		iatIndex++
```
leokondrashov, line 256: These two variables are the same now

leokondrashov (commented):
Some minor fixes in specification tests again. There is a change in the driver that is inconsistent with the rest of the code: experiment duration with second granularity. In the front-end we interpret experiment duration as seconds, but in driver, now, we try to interpret it as number of minutes. See the comment for functionsDriver.

pkg/driver/trace_driver.go
cvetkovic: Fixed.

pkg/generator/specification_test.go
```diff
@@ -78,167 +143,164 @@ func TestSerialGenerateIAT(t *testing.T) {
...
+			granularity:     common.MinuteGranularity,
+			expectedPoints: []float64{
+				13_794_305.088992357,
+				14_731_963.725122724,
+				11_882_919.652012857,
+			},
+			testDistribution: false,
+		},
+		{
+			testName:        "1inv_1min_exponential",
```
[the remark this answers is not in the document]
cvetkovic: Removed.

pkg/generator/specification_test.go
```diff
@@ -78,167 +143,164 @@ func TestSerialGenerateIAT(t *testing.T) {
...
-				},
+			expectedPoints: []float64{
+				0,
 			},
 			testDistribution: false,
 		},
 		{
-			testName:        "5min_5ipm_equidistant",
-			invocations:     []int{5, 5, 5, 5, 5},
+			testName:        "5inv_1min_equidistant",
```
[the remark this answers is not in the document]
cvetkovic: Removed.

pkg/generator/specification_test.go
```diff
@@ -78,167 +143,164 @@ func TestSerialGenerateIAT(t *testing.T) {
...
 			},
 			testDistribution: false,
 		},
 		{
-			testName:        "1min_25ipm_uniform_shift",
-			invocations:     []int{25},
-			iatDistribution: common.Uniform,
-			shiftIAT:        true,
+			testName:        "30inv_5min_equidistant",
+			invocations:     []int{5, 5, 5, 5, 5},
```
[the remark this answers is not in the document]
cvetkovic: Fixed.

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,417 @@
...
+						t.Errorf("Unexpected value fx %d val %d - got: %f; expected: %f", fIndex, i, matrix[fIndex][i], test.expectedIAT[fIndex][i])
+					}
+
+					if currentMinute > len(test.expectedCount[fIndex]) {
+						t.Errorf("Invalid expected count array size for function with index %d", fIndex)
+					}
+
+					if matrix[fIndex][i] >= 0 {
+						sum += matrix[fIndex][i]
+					}
```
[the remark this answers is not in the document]
cvetkovic: Removed.

pkg/driver/trace_driver.go
cvetkovic: Going with that approach then.

pkg/driver/trace_driver.go
cvetkovic: Replaced with `invocationCount := len(function.Specification.IAT)`

cvetkovic: I put `invocationCount := len(IAT)`. Seems that we were not following the convention in the tests on what we pass as input arguments. The `cmd/loader.go` operates as you describe, but tests we passing something else.

pkg/generator/rps_test.go
```diff
@@ -0,0 +1,412 @@
...
+
+			if len(matrix) != len(test.expectedIAT) {
+				t.Errorf("Unexpected number of functions - got: %d, expected: %d", len(matrix), len(test.expectedIAT))
+			}
+			if len(minuteCounts) != len(test.expectedCount) {
+				t.Errorf("Unexpected count array size - got: %d, expected: %d", len(minuteCounts), len(test.expectedCount))
+			}
+
+			for fIndex := 0; fIndex < len(matrix); fIndex++ {
+				currentMinute := 0
```
leokondrashov, line 388: This variable doesn't do anything

pkg/driver/trace_driver.go
```diff
@@ -236,177 +154,117 @@ func (d *Driver) invokeFunction(metadata *InvocationMetadata) {
...
-		}
-	}
-
-	*minuteIndex++
-	*invocationIndex = 0
-	*previousIATSum = 0
-
-	if d.Configuration.WithWarmup() && *minuteIndex == (d.Configuration.LoaderConfiguration.WarmupDuration+1) {
+func (d *Driver) announceWarmupEnd(start time.Time, currentPhase *common.ExperimentPhase) {
+	if *currentPhase == common.WarmupPhase && hasMinuteExpired(start) {
```
leokondrashov, line 261: This makes all warmup 1 minute long. I think we can safely revert back to the check for the current minute instead of real minute expiry (btw, I don't think that would work with second granularity and I don't see any particular use of it now).

leokondrashov (commented):
Please fix warmup handling. Looks fine to me, but I want to test-run it on several experiments to be sure that nothing goes wrong.

leokondrashov:
Unfortunately, I see a problem with warmup and RPS mode: (I modified the config file to include 1 minute of warmup)
```
$ go run cmd/loader.go -config cmd/config_knative_rps.json -verbosity debug
WARN[Dec  4 06:14:18.474] It is recommended that the first 10% of cold starts are discarded from the experiment results for low cold start RPS.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x999be7]

goroutine 1 [running]:
github.com/vhive-serverless/loader/pkg/trace.profileConcurrency(...)
        /users/lkondras/loader/pkg/trace/trace_profiler.go:87
github.com/vhive-serverless/loader/pkg/trace.DoStaticTraceProfiling({0xc0000908f0, 0x1, 0xc0001ffba8?})
        /users/lkondras/loader/pkg/trace/trace_profiler.go:38 +0x107
github.com/vhive-serverless/loader/pkg/driver.(*Driver).RunExperiment(0xc0000b9980)
        /users/lkondras/loader/pkg/driver/trace_driver.go:488 +0x38
main.runRPSMode(0xc00019cc60, 0x0, 0x0)
        /users/lkondras/loader/cmd/loader.go:236 +0x29e
main.main()
        /users/lkondras/loader/cmd/loader.go:111 +0x479
exit status 2
```

After closer inspection, I see that there are problems with RPS mode and warmup together: we generate only the experiment duration, but run for experiment duration + warmup.

I think the whole warmup might be broken now, previously, we were skipping one minute for profiling during driver execution, while now we do the invocation even for the minute that we were using for static profiling.

pkg/generator/rps_test.go
cvetkovic: Removed.

pkg/driver/trace_driver.go
cvetkovic: I am actually not sure about this. What about cases where you have 0 invocations during 5 minutes, and warmup was 1 minute. This function is warmed up at t=1min and scale is 0, but if we change to what you suggest, then warm-up minute will be the first minute in which function is invoked.

I would leave this as is, and extend `hasMinuteExpired` to expiration of `WarmupDuration * time.Second`.

leokondrashov: You can check it before the invocation happens so we will update the phase to record correctly. It might be written to the log a bit late (e.g., at the 5-minute mark in your example), but still, it will happen before the first invocation in the experiment phase.

cvetkovic: Fixed.

cvetkovic:
@leokondrashov: I think we should also change names of `ExperimentDuration` and `WarmupDuration` parameters and add `Minute` suffix, since it becomes unclear how this works if `Granularity` is `second`. It might be written in the doc, but we might have some bugs around this in the code now or in the future. What do you think?

leokondrashov:
I agree that it is extremely confusing. We should take care of that.

Right now, I've checked the loader once more, and it works as expected. I think we can merge this one and add an issue about the granularity. And work on the existing issues one-by-one so that the review process would be fast.

leokondrashov (approved):

cvetkovic:
Ok. I have created an issue. I will proceed with merging then.

PR #558, opened 2024-11-16
add loader dry run

nosnelmil:
## Summary

Adds a dry run flag to the loader, enabling users to execute the loader without generating trace loads or running experiments. Also adds a simple validation for cpu limit field in loader config.

* Add a dryRun flag to the loader which accepts a boolean value.
* Modifies the execution flow to exit early by checking the dryRun flag just before calling RunExperiment via the experiment driver in the runTraceMode function
* Add a simple validation to ensure provided cpu limit in loader config is valid

* N/A

* N/A

### DIFF
--- a/cmd/loader.go
+++ b/cmd/loader.go
@@ -27,10 +27,11 @@ package main
 import (
 	"flag"
 	"fmt"
-	"github.com/vhive-serverless/loader/pkg/generator"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/vhive-serverless/loader/pkg/generator"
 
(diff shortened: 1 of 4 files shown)


leokondrashov:
I think you forgot the CPU limit check that we were talking about, please add it

nosnelmil:
> I think you forgot the CPU limit check that we were talking about, please add it

Oh yes, thought I will be adding that in another PR, will add that in

leokondrashov:
LGTM, @cvetkovic please review as well

cvetkovic (approved):

PR #559, opened 2024-11-16
Feature/multi loader

nosnelmil:
## Summary

A wrapper around loader to run multiple studies at once with additional features like validation, dry-run and log collection

- **Multi-Loader Entry Point**:
  - `MultiLoaderRunner` houses the main logic of the multi-loader and exposes two execution points: `RunDryRun` and `RunActual`
  - `multi_loader.go` in `tools/multiloader/` initializes the `MultiLoaderRunner` and calls the appropriate execution point

- **Experiment Execution Flow**:
  - The multi-loader runner is instantiated with the provided configuration path
  - Execution is split into three phases: pre-execution, execution, and post-execution
    - **Pre-Execution**: Executes global pre-scripts, sets up directories, and unpacks sub-experiments.
    - **Execution**: Runs experiments with the generated configurations and appropriate flags and collect loader.go logs.
    - **Post-Execution**: Executes experiment-specific post-scripts and performs necessary cleanup tasks.
  - If the execution is a dry run, the same flow is followed, but the `dryRun` flag is set to `true` when calling `loader.go`

- **Post Execution**:
  - Global post-scripts are executed after the experiments finish
  - Executes `Make Clean`
  - Temporary folders are removed

* N/A

* N/A


leokondrashov (commented):
This is the first brief look at the code.

Please consider moving the multi-loader exclusive part away from the loader's common package. There is no point in storing a structure for a multi-loader there.

Another thing is gofmt; please set the VS Code to run it on all file saves. There are issues with alignment and whitespace around parentheses.

Please add GitHub workflows for the tests and add e2e tests. You can add separate workflow files for both. In e2e, please check the output hierarchy and the different settings used in the experiments.

Please review each function and ensure that the names make sense to someone who reads them for the first time.

For the failing spellcheck, you will need to add all the failing words (unless you misspelled them) into `.github/configs/wordlist.txt`.

tools/multi_loader/multi_loader.go
```diff
@@ -0,0 +1,62 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/vhive-serverless/loader/tools/multi_loader/runner"
+)
+
+var (
+	multiLoaderConfigPath = flag.String("multiLoaderConfigPath", "tools/multi_loader/multi_loader_config.json", "Path to multi loader configuration file")
+	verbosity             = flag.String("verbosity", "info", "Logging verbosity - choose from [info, debug, trace]")
+	iatGeneration         = flag.Bool("iatGeneration", false, "Generate iats only and skip invocations")
+	generated             = flag.Bool("generated", false, "If iats were already generated")
```
leokondrashov, line 16: Do we need both of them in both CLI args and config? I think one way to set them would be enough

tools/multi_loader/test_data/example_1_test/dirigent.csv
leokondrashov, line 1: I don't like the copy-paste of the same example trace repeatedly. If it is used in tests, we can create them from the original during the additional setup step. You can also add a script that does exactly this for the functionality demonstration cases.

tools/multi_loader/types/node_types.go
leokondrashov, line 1: This is also not in the current features, right? Please remove for now.

tools/multi_loader/runner/multi_loader_runner.go
```diff
@@ -0,0 +1,411 @@
...
+	var experiments []types.LoaderExperiment
+
+	// if user specified a trace directory
+	if study.TracesDir != "" {
+		experiments = d.unpackFromTraceDir(study)
+		// user define trace format and values instead of directory
+	} else if study.TracesFormat != "" && len(study.TraceValues) > 0 {
+		experiments = d.unpackFromTraceValues(study)
+	} else {
+		// Theres only one experiment in the study
```
leokondrashov, line 138: Comments are placed inconsistently (inside/outside, indented/not indented)

tools/multi_loader/runner/multi_loader_runner.go
```diff
@@ -0,0 +1,411 @@
...
+	// Run global prescript
+	common.RunScript(d.MultiLoaderConfig.PreScript)
+	// Iterate over studies and run them
+	for si, study := range d.MultiLoaderConfig.Studies {
+		log.Debug("Setting up study: ", study.Name)
+		// Run pre script
+		common.RunScript(study.PreScript)
+
+		// Unpack study to a list of studies with different loader configs
+		sparseExperiments := d.unpackStudy(study)
```
leokondrashov, line 88: What do you mean by sparse experiments?

tools/multi_loader/runner/multi_loader_runner.go
```diff
@@ -0,0 +1,411 @@
...
+
+	// Create the log file
+	logFilePath := path.Join(path.Dir(experiment.Config["OutputPathPrefix"].(string)), "loader.log")
+	logFile, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer logFile.Close()
+
+	for i := 0; i < NUM_OF_RETRIES; i++ {
```
leokondrashov, line 305: You can simplify the body a bit: compute the actual number of retries based on fail fast and dry run before the loop. In the body, on every iteration except the first, you log that it is retry; after that, you run execute the loader command and check for success, if so, return as completed. If we exit the loop, it means we failed, so move the handling code outside.

leokondrashov (commented):
Great job. I've made couple of comments about minor things, please fix

tools/multi_loader/runner/multi_loader_runner.go
nosnelmil: They contain partial laoder configs that would be merged with the base config later. So partial should be a better word, sparse was the first thing that came to mind. Will make the changes

tools/multi_loader/runner/multi_loader_runner.go
nosnelmil: That's a good point! Will make the update. Also currently, NUM_OF_RETIRES is used as the total number of executions which is confusing will update the logic for that too.

tools/multi_loader/multi_loader.go
nosnelmil: The purpose of the CLI arguments is to allow users to set these flags at a global level (applying to all studies). Having these flags as fields in the config is to give users the ability to override their values for a specific study (enabling study-level control)

nosnelmil: Instead of having users define both global and study level flags in the config, i made the global level flags as cli args to make it consistent with loader.go

tools/multi_loader/test_data/example_1_test/dirigent.csv
nosnelmil: Thats true, will work on it

tools/multi_loader/common/utils.go
```diff
@@ -0,0 +1,52 @@
...
+		log.Fatal(err)
+	}
+
+	err = os.WriteFile(path, configByteValue, 0644)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func DeterminePlatformFromConfig(multiLoaderConfig types.MultiLoaderConfiguration) string {
```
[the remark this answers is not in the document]
nosnelmil: It should work if the necessary fields are present in the study config. But, since there are different base configs for various platforms, saying that this supports multiple platforms is not that right. I could add support for multiple platforms if there is a use case for it, but for now I will add a note about this in the docs

tools/multi_loader/multi_loader.go
leokondrashov: I don't think we need to make it consistent between the two. Especially since these options won't work properly with a sequence of experiments: they read/write in the same directory you run, so it might read the wrong or overwrite IAT files.

tools/multi_loader/runner/multi_loader_runner.go
```diff
@@ -0,0 +1,451 @@
...
+		if err := d.executeLoaderCommand(experiment, logFile); err != nil {
+			log.Error(err)
+			log.Error("Experiment failed: ", experiment.Name)
+			logFile.WriteString("Experiment failed: " + experiment.Name + ". Error: " + err.Error() + "\n")
+			// Log retry attmempt if necessary
+			if i == 0 {
+				log.Info("Retrying experiment ", experiment.Name)
+				logFile.WriteString("==================================RETRYING==================================\n")
+				experiment.Verbosity = "debug"
+			}
```
leokondrashov, line 347: This should be printed in any of the retries (I wouldn't rely on the `NUM_OF_RETRIES` being one; we might change it based on the task). I would propose putting logging at the beginning of the body under `i != 0` condition. Change verbosity after the fail (doesn't need to check for the first failure or not).

tools/multi_loader/runner/multi_loader_runner.go
```diff
@@ -0,0 +1,451 @@
...
+	log.Debug("Runnning Cleanup")
+	// Run make clean
+	if err := exec.Command("make", "clean").Run(); err != nil {
+		log.Error("Error occured while running cleanup", err)
+	}
+	// Remove temp file
+	os.Remove(EXPERIMENT_TEMP_CONFIG_PATH)
+
+	log.Debug("Cleanup completed")
+}
```
leokondrashov, line 451: Actually, this doesn't need to be run. Main part of clean up is done by loader (the un-deployment of functions). Also, most of the commands there are knative-related and cleaning Prometheus, which not everyone needs (sometimes the opposite, you might want to retain the history there). I'd propose to remove the step. If needed, it can be added in a postscript.

tools/multi_loader/runner/multi_loader_runner_test.go
```diff
@@ -0,0 +1,276 @@
...
+		cleanup := setup()
+		defer cleanup()
+		// Create a new multi-loader driver with the test config path
+		multiLoader, err := NewMultiLoaderRunner(multiLoaderTestConfigPath, "info", false, false, false)
+		if err != nil {
+			t.Fatalf("Failed to create multi-loader driver: %v", err)
+		}
+		funcToTest(t, *multiLoader)
+	}
+}
```
leokondrashov, line 204: Looks quite complicated. Can you replace it with a single setup function to create the entire environment and clean up everything left after the previous experiment? It creates two levels of indentation to insert cleanup after the experiment, which could be run as a part of the next experiment.

tools/multi_loader/runner/multi_loader_runner_test.go
```diff
@@ -0,0 +1,276 @@
...
+	originalExitFunc := log.StandardLogger().ExitFunc
+
+	// replace logrus exit function
+	log.StandardLogger().ExitFunc = func(int) { fatal = true }
+
+	funcToTest()
+	// restore original state
+	log.StandardLogger().ExitFunc = originalExitFunc
+	assert.True(t, fatal, "Expected log.Fatal to be called")
+}
```
leokondrashov, line 276: This prints the fatal messages into the test logs, which looks confusing (fatal put test passes). Is there a way to insert something in the log that this is expected? I think right here, before running the test, add the log message: `Expecting a fatal message during the test, overriding the exit`.

tools/multi_loader/runner/multi_loader_runner.go
nosnelmil: That's a good point, but I would still have to clean up temp resources created by multi loader. Having the cleanup function also helps to structure the flow of multi-loader clearly so I would suggest keeping the function but removing the execution of make clean

tools/multi_loader/runner/multi_loader_runner_test.go
nosnelmil: Yes the structure could be better due to the double indentation required, but my intention was to prevent having the need to remember calling certain functions when creating tests. I will make the changes accordingly

leokondrashov (commented):
LGTM

@cvetkovic please take a look. This is a wrapper tool for the loader. It should replace derelict experiment driver that we have lingering and confusing people. We also have several usability features coming up, like parameter sweep (useful for RPS mode), top measurements, and knative logs collection. If you have other enhancements in mind, please let us know.

pkg/common/validators.go
```diff
@@ -1,12 +1,45 @@
...
+	// -oStrictHostKeyChecking=no -p 22
+	out, err := cmd.CombinedOutput()
+	if bytes.Contains(out, []byte("Permission denied")) || err != nil {
+		log.Error(string(out))
+		log.Fatal("Failed to connect to node ", node)
+	}
+}
+
+func CheckPath(path string) {
+	if (path) == "" {
```
cvetkovic, line 27: No need for brackets.

cvetkovic:
LGTM!

leokondrashov:
Good. Then, please squash commits (at least to less than 10 if you want to separate some of the contributions), fix the comment by Lazar and we are ready to merge.

leokondrashov (approved):

PR #567, opened 2024-12-06
RPS mode data size feature

cvetkovic:


leokondrashov:
I see this data size is only applicable in RPS and HTTP invoker.

So, a couple of questions. Do we have or can we add comparable data movement for trace execution? The other question would be to find some dataset to make it realistic.

And what about RPC mode? We have a message in protobuf that we don't use at all; we can use it to model data movement.

cvetkovic:
@leokondrashov: There is an opportunity to add data movement for trace execution. [This paper](https://dl.acm.org/doi/10.1145/3472883.3486974) gives some insight into distribution and releases the trace. I think this could be nice project for a student to take care of -- look into the data, find a way to match functions in the execution trace with data trace, and add such a feature in loader.

Regarding the RPC mode, I can add support for it in gRPC. I think this will perform poorly with large data payloads (like 1MB and bigger), so I expect significant overhead there. What do you think?

leokondrashov:
The payloads are quite small in this paper, up to 10s KB in the majority of cases, so I don't see a problem with adding it to gRPC.

The other question is also how to interpret these numbers, they are not only about inputs that we send directly with the requests, but also outputs. Originally, it was intended more as a storage workload characterization, not data moved through HTTP with request/response.

Previously, @JooyoungPark73 was doing some work on that front with the Azure Blob trace you mentioned. I think we should have some discussion about that.

PR #569, opened 2024-12-19
Integrate Azure Functions deployment with In-Vitro

cavinkavi:
## Summary

To integrate Azure Functions deployment with In-Vitro. This feature enables deployment of serverless functions on the Azure Functions platform.

This PR adds a workflow to deploy multiple functions to Azure Functions via ZIP deployment.

Create an Azure Service Principal: `az ad sp create-for-rbac --name "InVitro" --role Contributor --scopes /subscriptions/<your-subscription-id>`

Login with Service Principal credentials: `az login --service-principal --username <appId> --password <password> --tenant <tenant-id>`

To start deployment: `go run cmd/loader.go --config cmd/config_azure_trace.json`

* `azurefunctions_setup`: this folder contains the relevant config and workload files to be used for deployment.
* `azure_functions.go`: initializes Azure resources required for deployment to Azure Functions, such as Resource Group, Storage Account, and Function App. In order to deploy multiple functions via ZIP, each function must have its own specific folder containing required dependencies, before zipping and deploying all the function folders as a single package. After successful deployment, local temporary folders holding the functions and the zip package are cleaned up. 
(Note: Azure resource cleanup (e.g., Resource Group) must be performed manually in the Azure Portal for now. This will be addressed in future PRs focused on invocation logic. All cleanup, including local temporary folders, will be done after entire experiment is completed.)

* gopkg.in/yaml.v2 for reading Azure configuration files.
* Azure CLI for managing Azure resources and deployment.

* N/A

### DIFF
--- a/azurefunctions_setup/shared_azure_workload/azureworkload.py
+++ b/azurefunctions_setup/shared_azure_workload/azureworkload.py
@@ -1,0 +1,71 @@
+import time
+import os
+import socket
+import json
+from typing import Dict
+
+# Global variables for IterationsMultiplier and hostname
+IterationsMultiplier = 102  # Assuming similar cloud benchmark setup
+hostname = socket.gethostname()
+
+# Simulate the busySpin function
+def busy_spin(runtime_milli: int):
+    total_iterations = IterationsMultiplier * runtime_milli
+    for _ in range(total_iterations):
+        # Simulate a math-heavy operation
+        sqrt_of_10 = 10 ** 0.5
+
+# Convert TraceFunctionExecution
+def trace_function_execution(start: float, time_left_milliseconds: int) -> str:
+    time_consumed_milliseconds = int((time.time() - start) * 1000)
+    if time_consumed_milliseconds < time_left_milliseconds:
+        time_left_milliseconds -= time_consumed_milliseconds
+        if time_left_milliseconds > 0:
+            busy_spin(time_left_milliseconds)
+
+    return f"OK - {hostname}"
+
+# The handler function for Azure Functions (Python)
+import azure.functions as func
+import logging
+
+def main(req: func.HttpRequest) -> func.HttpResponse:
+    logging.info("Processing request.")
+
+    start_time = time.time()
+
+    # Parse JSON request body
+    try:
+        req_body = req.get_json()
+        logging.info(f"Request body: {req_body}")
+    except ValueError:
+        logging.error("Invalid JSON received.")
+        return func.HttpResponse(
+            json.dumps({"error": "Invalid JSON"}),
+            status_code=400,
+            mimetype="application/json"
+        )
+
+    runtime_milliseconds = req_body.get('RuntimeInMilliSec', 1000)
+    memory_mebibytes = req_body.get('MemoryInMebiBytes', 128)
+
+    logging.info(f"Runtime requested: {runtime_milliseconds} ms, Memory: {memory_mebibytes} MiB")
+
+    # Trace the function execution (busy work simulation)
+    result_msg = trace_function_execution(start_time, runtime_milliseconds)
+
+    # Prepare the response
+    response = {
+        "DurationInMicroSec": int((time.time() - start_time) * 1_000_000),
+        "MemoryUsageInKb": memory_mebibytes * 1024,
+        "Message": result_msg
+    }
+
+    logging.info(f"Response: {response}")
+
+    return func.HttpResponse(
+        json.dumps(response),
+        status_code=200,
+        mimetype="application/json"
+    )
+
--- a/pkg/driver/deployment/azure_functions.go
+++ b/pkg/driver/deployment/azure_functions.go
@@ -1,0 +1,321 @@
+package deployment
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/vhive-serverless/loader/pkg/common"
+	"github.com/vhive-serverless/loader/pkg/config"
+	"gopkg.in/yaml.v2"
+)
+
+// Config struct to hold Azure Function deployment configuration
+type Config struct {
+	AzureConfig struct {
+		ResourceGroup      string `yaml:"resource_group"`
+		StorageAccountName string `yaml:"storage_account_name"`
+		FunctionAppName    string `yaml:"function_app_name"`
+		Location           string `yaml:"location"`
+	} `yaml:"azurefunctionsconfig"`
+}
+
+type azureFunctionsDeployer struct {
+	functions []*common.Function
+}
+
+func newAzureFunctionsDeployer() *azureFunctionsDeployer {
+	return &azureFunctionsDeployer{}
+}
+
+func (afd *azureFunctionsDeployer) Deploy(cfg *config.Configuration) {
+	afd.functions = cfg.Functions
+	deployAzureFunctions(afd.functions)
+}
+
+func (afd *azureFunctionsDeployer) Clean() {
+	cleanAzureFunctions(afd.functions)
+}
+
+func deployAzureFunctions(functions []*common.Function) {
+
+	// 1. Initialize resources required for Azure Functions deployment
+	// 2. Create function folders
+	// 3. Zip function folders
+	// 4. Deploy the function to Azure Functions
+
+	// Load azurefunctionsconfig yaml file
+	config, err := LoadConfig("azurefunctions_setup/azurefunctionsconfig.yaml")
+	if err != nil {
+		log.Fatalf("Error loading azure functions config yaml: %s", err)
+	}
+
+	baseDir := "azure_functions_for_zip"
+
+	// 1. Initialize resources required for Azure Functions deployment
+	initAzureFunctions(config)
+
+	// 2. Create function folders
+	if err := createFunctionFolders(baseDir, functions); err != nil {
+		log.Fatalf("Error setting up function folders required for zipping: %s", err)
+	}
+
+	// 3. Zip function folders
+	if err := ZipFunctionAppFiles(); err != nil {
+		log.Fatalf("Error zipping function app files for deployment: %s", err)
+	}
+
+	// 4. Deploy the function to Azure Functions
+	if err := DeployFunction(config, functions); err != nil {
+		log.Fatalf("Error deploying function: %s", err)
+	}
+
+}
+
+func cleanAzureFunctions(functions []*common.Function) {
+
+	//Delete created folders
+	//Delete zip file
+	//Delete Azure resources
+
+}
+
+/* Functions for initializing resources required for Azure Functions deployment */
+
+func initAzureFunctions(config *Config) {
+
+	// 1.Create Resource Group
+	// 2.Create Storage Account
+	// 3.Create Function App
+	// 4.Set WEBSITE_RUN_FROM_PACKAGE
+
+	//checkDependencies()	ToDo: Check if all required dependencies are installed on VM
+
+	// 1. Create Resource Group
+	if err := CreateResourceGroup(config); err != nil {
+		log.Fatalf("Error during Resource Group creation: %s", err)
+	}
+
+	// 2. Create Storage Account
+	if err := CreateStorageAccount(config); err != nil {
+		log.Fatalf("Error during Storage Account creation: %s", err)
+	}
+
+	// 3. Create Function App
+	if err := CreateFunctionApp(config); err != nil {
+		log.Fatalf("Error during Function App creation: %s", err)
+	}
+
+	// 4. Set WEBSITE_RUN_FROM_PACKAGE
+	if err := SetWebsiteRunFromPackage(config); err != nil {
+		log.Fatalf("Error setting WEBSITE_RUN_FROM_PACKAGE: %s", err)
+	}
+
+	log.Info("Azure Functions environment for deployment initialized successfully.")
+}
+
+// LoadConfig reads the YAML configuration file
+func LoadConfig(filePath string) (*Config, error) {
+	config := &Config{}
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+	err = yaml.Unmarshal(data, config)
+	return config, err
+}
+
+// CreateResourceGroup creates an Azure Resource Group: {az group create --name <resource-group> --location <location>}
+func CreateResourceGroup(config *Config) error {
+	createResourceGroupCmd := exec.Command("az", "group", "create",
+		"--name", config.AzureConfig.ResourceGroup,
+		"--location", config.AzureConfig.Location)
+
+	if err := createResourceGroupCmd.Run(); err != nil {
+		return fmt.Errorf("failed to create resource group: %w", err)
+	}
+
+	log.Infof("Resource group %s created successfully.", config.AzureConfig.ResourceGroup)
+	return nil
+}
+
+// CreateStorageAccount creates an Azure Storage Account : {az storage account create --name <storage-account-name> --resource-group <resource-group> --location <location> --sku Standard_LRS}
+func CreateStorageAccount(config *Config) error {
+	cmd := exec.Command("az", "storage", "account", "create",
+		"--name", config.AzureConfig.StorageAccountName,
+		"--resource-group", config.AzureConfig.ResourceGroup,
+		"--location", config.AzureConfig.Location,
+		"--sku", "Standard_LRS")
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to create storage account: %w", err)
+	}
+
+	log.Infof("Storage account %s created successfully.", config.AzureConfig.StorageAccountName)
+	return nil
+}
+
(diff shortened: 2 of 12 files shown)


azurefunctions_setup/shared_azure_workload/azureworkload.py
```diff
@@ -0,0 +1,71 @@
+import time
+import os
+import socket
+import json
+from typing import Dict
```
wanghanchengchn, line 5: Could we review if the imports os and Dict are necessary? From what I can see, we're not using any functionality from the os module, nor any Dict type annotations in the code.

azurefunctions_setup/shared_azure_workload/azureworkload.py
```diff
@@ -0,0 +1,71 @@
...
+    if time_consumed_milliseconds < time_left_milliseconds:
+        time_left_milliseconds -= time_consumed_milliseconds
+        if time_left_milliseconds > 0:
+            busy_spin(time_left_milliseconds)
+
+    return f"OK - {hostname}"
+
+# The handler function for Azure Functions (Python)
+import azure.functions as func
+import logging
```
wanghanchengchn, line 30: For better readability and adherence to PEP 8, would you mind moving the imports on lines 29-30 (import azure.functions as func and import logging) to the top of the file with the other imports? This helps maintain a consistent structure where all imports are grouped at the beginning of the file.

azurefunctions_setup/shared_azure_workload/azureworkload.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Convert TraceFunctionExecution
+def trace_function_execution(start: float, time_left_milliseconds: int) -> str:
+    time_consumed_milliseconds = int((time.time() - start) * 1000)
+    if time_consumed_milliseconds < time_left_milliseconds:
+        time_left_milliseconds -= time_consumed_milliseconds
+        if time_left_milliseconds > 0:
+            busy_spin(time_left_milliseconds)
+
+    return f"OK - {hostname}"
```
wanghanchengchn, line 26: 1. About `IterationsMultiplier = 102`:
  - How was this number determined?
  - Is it environment-specific?
  - Is there any documentation or benchmark results we can reference?

2. Could we use `time.sleep()` here? I guess this code aims to keep CPU busy rather than just waiting. I assume we can't use `time.sleep()` here because it would make CPU idle and affect monitoring metrics. Is my understanding correct?

3. If we need to maintain CPU activity, could we simplify the busy spin implementation to:
  ```python
  def busy_spin(duration_ms: int) -> None:
      end_time = perf_counter() + duration_ms/1000
      while perf_counter() < end_time:
          continue
  ```
This seems to achieve the same goal with cleaner logic while still keeping CPU active for monitoring?

pkg/driver/deployment/azure_functions.go
```diff
@@ -0,0 +1,321 @@
...
+	// 2.Create Storage Account
+	// 3.Create Function App
+	// 4.Set WEBSITE_RUN_FROM_PACKAGE
+
+	//checkDependencies()	ToDo: Check if all required dependencies are installed on VM
+
+	// 1. Create Resource Group
+	if err := CreateResourceGroup(config); err != nil {
+		log.Fatalf("Error during Resource Group creation: %s", err)
+	}
```
wanghanchengchn, line 101: I noticed that we're missing some important prerequisite documentation that would help users successfully run Azure CLI commands.

Could you please add some documentation to explain what needs to be done before running the program? For example, how to install Azure CLI, what needs to be done after installing Azure CLI, such as running az login, and what else needs to be done afterward?

This documentation would be very helpful for users trying to run the code for the first time. Feel free to reference our existing documentation structure for consistency.

pkg/driver/deployment/azure_functions.go
```diff
@@ -0,0 +1,321 @@
...
+	// Call the cleanup function after deployment, to delete temp folders and files
+	if err := cleanUpDeploymentFiles("azure_functions_for_zip", "azurefunctions.zip"); err != nil {
+		log.Errorf("Error during cleanup: %s", err)
+	} else {
+		log.Info("Deployment and cleanup of zip files completed successfully.")
+	}
+
+	//Stop the program after deployment for testing purposes, to remove after invocation is implemented
+	log.Info("Stopping program after deployment phase for PR.")
+	os.Exit(0) // Exit with status code 0 (successful execution)
```
wanghanchengchn, line 298: What is the purpose of this code block? Is it really necessary? From my understanding, this kind of code shouldn't appear in a production environment. Could you please remove this forceful program termination? Using os.Exit(0) prevents proper cleanup operations and makes the subsequent return nil unreachable. It would be better to let the program complete its execution naturally.

pkg/driver/deployment/azure_functions.go
```diff
@@ -0,0 +1,321 @@
...
+
+}
+
+func cleanAzureFunctions(functions []*common.Function) {
+
+	//Delete created folders
+	//Delete zip file
+	//Delete Azure resources
+
+}
```
wanghanchengchn, line 85: Could you please add a space after // in the comments to maintain consistent code formatting? This helps maintain code style consistency throughout the project.

azurefunctions_setup/shared_azure_workload/azureworkload.py
cavinkavi: Changes made as suggested.

azurefunctions_setup/shared_azure_workload/azureworkload.py
cavinkavi: Removed `IterationsMultiplier=102` and updated busy spin implementation to use `perf_counter()`.

pkg/driver/deployment/azure_functions.go
cavinkavi: Changes made as suggested.

pkg/driver/deployment/azure_functions.go
cavinkavi: Changes made as suggested. Also, updated `cleanAzureFunctions` to perform clean up after experiment.

wanghanchengchn:
Looks like there are some errors in the checks.

1. Missing commit signatures: It seems the commits are missing `Signed-off-by` signatures. Please use `git commit -s -m "Your commit message"` when committing.

2. Commit history: Consider squashing your commits into a single commit to maintain a clean history.

You can find more details about these requirements in our contributing guide:
https://github.com/vhive-serverless/vHive/blob/main/docs/contributing_to_vhive.md

wanghanchengchn:
Hi @leokondrashov, I've completed my initial review of this PR. @cavinkavi has made the corresponding changes, and the current version looks good to me. Do you have any additional feedback or concerns?

azurefunctions_setup/shared_azure_workload/azureworkload.py
```diff
@@ -0,0 +1,71 @@
...
```
leokondrashov, line 25: Please reuse this parts from existing Python implementation (`server/trace-func-py/trace_func.py`). Don't reimplement because this creates duplicates in the code and incorrect performance comparisons.

leokondrashov (commented):
First, there are no tests for the infrastructure, please add. Better even add them into CI tests.
Second, does Azure support Go deployments? We don't use Python that much and I'm not sure that our main Python implementation is well-behaved. It would be preferable to use Go instead.

leokondrashov:
No need to throw out the Python deployment, though. It could be an option to deploy one of them during runtime. But the default should be Go to align with other deployments.

pkg/driver/deployment/azure_functions_test.go
leokondrashov, line 1: Here you have the end-to-end test. Please add a test that would check the potential zip file for being operational. This wouldn't require the Azure credentials, so we can run it without any concerns about spending. This test would show when we have problems with our file copies in the archive.

Please also add them to the CI, so we can see the results right in the PRs.

At the same time, we can have an end-to-end test by just running the `loader.go` with the correct config file and checking for the output. There is no reason to add the script that duplicates the experiment pipeline. Better to repurpose it to test specific parts of the pipeline (deploying/cleaning functions, checking for the zip health).

leokondrashov (commented):
Also, please change the message of your commit to something meaningful. Addressing my comments doesn't make any sense when you are looking at the history of commits.

pkg/driver/deployment/azure_functions_test.go
cavinkavi: In that case, would it be better if I remove the azure_functions_test.go file and create an End-to-End test under workflows, which will run the go command and check the output as you have mentioned?

leokondrashov: Yes, similar to existing e2e tests. But still, add the module tests that I mentioned. For them, you would need the *_test.go files. And don't forget to add them into CI.

azurefunctions_setup/extract-workload.py
```diff
@@ -0,0 +1,66 @@
...
+        content = f.read()
+
+    # Use regex to extract the execute_function definition and body
+    match = re.search(r"def execute_function\(.*?\):.*?(?=def |\Z)", content, re.DOTALL)
+    if not match:
+        raise ValueError("execute_function() not found in trace_func.py")
+
+    return match.group(0)
+
+def inject_function(func_code, workload_path):
```
leokondrashov, line 22: I don't know how this slipped from the previous review.

Please don't look inside the file and parse it to extract the implementation during deployment. What I proposed is to pull this function into a separate file and reuse it from both places. This way, during deployment, you only need to copy the file and import. I don't trust parsing the files, regexp are not as robust as you would think.

docs/loader.md
```diff
@@ -262,4 +262,43 @@ Note:
...
+**Quick Setup for Azure Deployment:**
+1. Install the Azure CLI and verify installation:
+    ```bash
+    curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+    az --version
+    ```
+2. Log in as a user (Note: This will open a browser window to select Azure account):
+    ```bash
+    az login
+    ```
```
leokondrashov, line 283: This is not always possible. There are no browser available on cloudlab nodes. Can you provide pure CLI method of logging in?

docs/loader.md
```diff
@@ -262,4 +262,43 @@ Note:
...
+   ```bash
+   az login --service-principal --username $AZURE_APP_ID --password $AZURE_PASSWORD --tenant $AZURE_TENANT
+   ```
+6. Start the Azure Functions deployment experiment:
+    ```bash
+    go run cmd/loader.go --config cmd/config_azure_trace.json
+    ```
+---
+Note:
+- Current deployment is via ZIP
```
leokondrashov, line 304: Please add the note that Go is unsupported in Consumption Plan (or whatever we use) and we are using Python instead

leokondrashov (commented):
And still, the commit name doesn't make sense. In this PR, you introduce the Azure Function deployment. Please put this as the first line of the commit message (and the only, I don't see the reason to put any other comments there apart from sign-offs, feel free to disagree).

azurefunctions_setup/shared_azure_workload/exec_func.py
leokondrashov, line 1: It is exactly the same as in the server/trace-func-py. Please leave only one of them.

And add the other one in .gitignore file so we won't see it added in `git status` after we do deployment. Also add the note in the main py file for Azure deployment near the execute_function call about the fact that you need to copy it for local development.

docs/loader.md
```diff
@@ -262,4 +262,48 @@ Note:
...
+    ```
+   > Refer to Note section for generation of Service Principal credentials
+3. Start the Azure Functions deployment experiment:
+    ```bash
+    go run cmd/loader.go --config cmd/config_azure_trace.json
+    ```
+---
+Notes:
+
+- Service Principal must be created before running experiment, as some environments do not have browsers (e.g. CloudLab). Perform these steps in an environment that allows launching of browser and use the generated credentials.
```
leokondrashov, line 293:
```suggestion
- Service Principal must be created before running the experiment, as some environments do not GUI (e.g., CloudLab nodes). You can perform these steps in an environment that allows the launching of the browser and use the generated credentials. 
```

Overall, can this be done via Azure web console? This is much simpler than doing it via CLI, since this part you don't need to do in experiment environment.

.github/workflows/unit-tests-azure.yaml
leokondrashov, line 1: Instead of whole this file, you can add the directory into modules to test in `unit-tests.yaml`. Either way, you run all of the tests that you wrote in azure_functions_test.go but you specify all of them one-by-one, which is prone to errors (add test, forgot to test or the opposite, it fails because you removed one).

pkg/driver/deployment/azure_functions_test.go
```diff
@@ -0,0 +1,226 @@
...
+// mockExecCommand simulates exec.Command for testing
+func mockExecCommand(output string, err error) deployment.CommandRunner {
+	return func(name string, arg ...string) *exec.Cmd {
+		cmd := exec.Command("echo", output) // Simulate success
+		if err != nil {
+			return exec.Command("false") // Simulate failure
+		}
+		return cmd
+	}
+}
```
leokondrashov, line 28: What does this thing do? I only see it used once with `err=nil`. It overloads the code with unnecessary flexibility.

pkg/driver/deployment/azure_functions_test.go
```diff
@@ -0,0 +1,226 @@
...
+	assert.Equal(t, "https://test-app.azurewebsites.net/api/function0", functions[0].Endpoint)
+	assert.Equal(t, "https://test-app.azurewebsites.net/api/function1", functions[1].Endpoint)
+
+	// Check log output
+	logs := logBuffer.String()
+	assert.Contains(t, logs, "Deploying 2 functions to Azure Function App")
+	assert.Contains(t, logs, "Deployed all 2 functions successfully")
+}
+
+// Tests CleanUpDeploymentFiles function correctly removes the specified temporary files and directories.
```
leokondrashov, line 200: This test tests just two rm commands. It seems excessive for me. When I was talking about cleanup I meant the cleaning up the deployment in Azure.

pkg/driver/deployment/azure_functions_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+	// Cleanup: Remove the created files and directories after the test
+	err = os.RemoveAll(baseDir)
+	assert.NoError(t, err, "Failed to remove temp directory")
+	err = os.Remove(zipFilePath)
+	assert.NoError(t, err, "Failed to remove temp zip file")
+}
+
+// Tests DeployFunction function by simulating deployment of zip file to Azure.
+func TestDeployFunction(t *testing.T) {
```
leokondrashov, line 155: This test just deploys the functions without cleaning them. So, my guess is that it fails on the second time onwards because functions already exist. Please test cleaning them up in the same test, so we can see whether it works. Also, be careful with the collision of tests. I.e., you shouldn't mistakenly remove the e2e functions that happen to run at the same time.

pkg/driver/deployment/azure_functions.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	log.Debugf("Created %d function folders with copies of azureworkload.py, exec_func.py and function.json under %s folder.\n", len(function), baseDir)
+	return nil
+}
+
+/* Functions for zipping created function folders */
+
+func ZipFunctionAppFiles() error {
+	// Use bash to zip the contents of azure_functions_for_zip/ along with host.json and requirements.txt
+	cmd := exec.Command("bash", "-c", "cd azure_functions_for_zip && zip -r ../azurefunctions.zip . && cd .. && zip -j azurefunctions.zip azurefunctions_setup/host.json azurefunctions_setup/requirements.txt")
```
leokondrashov, line 279: Please add all of the temporary directories and files into .gitignore

pkg/driver/deployment/azure_functions.go
```diff
@@ -0,0 +1,353 @@
...
+
+func DeployFunction(config *Config, function []*common.Function, runner CommandRunner) error {
+	log.Infof("Deploying %d functions to Azure Function App...", len(function))
+
+	// Path to the zip file that contains the Python binary and other resources for deployment to Azure
+	zipFilePath := "azurefunctions.zip"
+
+	// Deploy the zip file to Azure Function App using CLI
+	cmd := runner("az", "functionapp", "deployment", "source", "config-zip",
+		"--name", config.AzureConfig.FunctionAppName,
```
leokondrashov, line 298: There is an issue with deploying all functions into the same application. In Azure, the scaling unit is an application, so when you invoke one of the functions, it will instantiate all of the functions in the app. So, if you call other functions in the application, it would be a warm start instead of a cold start. We should model independent scaling for all functions, so they should be in different applications.

pkg/driver/deployment/azure_functions.go
```diff
@@ -0,0 +1,353 @@
...
+		return fmt.Errorf("failed to zip function app files for deployment: %w", err)
+	}
+
+	log.Debug("Functions for deployment zipped successfully.")
+	return nil
+}
+
+/* Functions for deploying zipped functions */
+
+func DeployFunction(config *Config, function []*common.Function, runner CommandRunner) error {
```
leokondrashov, line 290: BTW, this name is confusing, it deploys all the functions, not just one

PR #582, opened 2025-01-17
Extract RPS knob from Platform setting

### LINKED ISSUE #536: Separate RPS mode option from platform
## Version & Branch

main branch

## Expected Behaviour

In the configuration of the loader, to enable RPS mode, we need to append `-RPS` to the platform option, which doesn't make much sense. To make it sensible, we better use a separate knob for the RPS mode or replace the trace with a special value (since we can have either trace mode or RPS mode).

leokondrashov:
## Summary

Fixes #536. Extracts the RPS option from Platform. 

* Create a special value for the trace selector (`RPS`). This would work because we can use either trace path or RPS mode, but not both.

* N/A

* Changes the Platform and TracePath configuration parameters.

### DIFF
--- a/cmd/loader.go
+++ b/cmd/loader.go
@@ -28,7 +28,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/vhive-serverless/loader/pkg/generator"
@@ -93,14 +92,9 @@ func main() {
 
 	supportedPlatforms := []string{
 		"Knative",
-		"Knative-RPS",
 		"OpenWhisk",
-		"OpenWhisk-RPS",
 		"AWSLambda",
-		"AWSLambda-RPS",
 		"Dirigent",
-		"Dirigent-RPS",
-		"Dirigent-Dandelion-RPS",
 		"Dirigent-Dandelion",
 	}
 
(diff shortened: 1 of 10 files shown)


cmd/loader.go
```diff
@@ -93,29 +92,24 @@ func main() {
...
-		"Dirigent-Dandelion-RPS",
 		"Dirigent-Dandelion",
 	}
 
 	if !slices.Contains(supportedPlatforms, cfg.Platform) {
 		log.Fatal("Unsupported platform!")
 	}
 
-	if cfg.Platform == "Knative" || cfg.Platform == "Knative-RPS" {
+	if cfg.Platform == "Knative" {
```
cvetkovic, line 105: I would compare lowercase strings just to be sure.

cmd/loader.go
```diff
@@ -93,29 +92,24 @@ func main() {
...
 		common.CheckCPULimit(cfg.CPULimit)
 	}
 
-	if !strings.HasSuffix(cfg.Platform, "-RPS") {
-		runTraceMode(&cfg, *iatFromFile, *iatGeneration)
-	} else {
+	if cfg.TracePath == "RPS" {
```
cvetkovic, line 109: Same here.

cvetkovic (approved):

PR #585, opened 2025-01-30
Tweaked Driver to invoke vSwarm functions

aryans1204:
## Summary

A small summary of the requirements (in one/two sentences).

## Implementation Notes :hammer_and_pick:

* Briefly outline the overall technical solution. If necessary, identify talking points where the reviewer's attention should be drawn to.
Made changes to `pkg/trace/` and `pkg/driver/` and `cmd/loader.go` to support execution of vSwarm functions and integrate with downstream vSwarm invoker.

## External Dependencies :four_leaf_clover:

* 

## Breaking API Changes :warning:

* 

*Simply specify none (N/A) if not applicable.*

PR #605, opened 2025-02-12
Added RPS support for dirigent-dandelion-workflows

tobystocki:
## Summary

Adds RPS support for dirigent-dandelion-workflows.

- We have extended the dirigent cluster manager to be able to orchestrate workflows and want to benchmark it in order to further optimize it. So orchestration is done within dirigent and the workflows are to be provided when registering them in the system. From outside the system registering workflows is different to registering a single function but invoking either one looks the same to the user.
- The workflow description is exported as a file that is read and the content is given to dirigent in the registration. The structure is not platform specific (though only works on dirigent) but as already said the orchestration inside dirigent only supports the dandelion platform.
- In this PR we extend the dirigent deployer and HTTP client to be able to also handle workflow requests beside the function requests it already can. This requires a few more configuration parameters.
- Moves dirigent specific configuration parameters into separate dirigent config file that is loaded only when the platform is Dirigent. This also affects the backend which is now a dirigent config parameter and no longer a separate platform.
- Adapts the changes from [PR #606](https://github.com/vhive-serverless/invitro/pull/606/files).

* _none_

* _none_


cvetkovic (changes requested):
Looks good. Just some minor comments above.

cmd/loader.go
```diff
@@ -151,7 +152,8 @@ func parseYAMLSpecification(cfg *config.LoaderConfiguration) string {
 	case "firecracker":
 		return "workloads/firecracker/trace_func_go.yaml"
 	default:
-		if cfg.Platform != "Dirigent" && cfg.Platform != "Dirigent-Dandelion" {
+		if cfg.Platform != "Dirigent" && cfg.Platform != "Dirigent-Dandelion" &&
```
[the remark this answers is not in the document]
tobystocki: I might be confusing this but as far as I understood the DAG mode generates DAG workflows dynamically and this is not what we are doing here since we are submitting predefined workflows (the queries). That's why I chose to create another platform type "dirigent-dandelion-workflow".
What do you @leokondrashov think about this?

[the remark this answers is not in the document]
leokondrashov: If you ask an external for comment, please provide proper explanation what and how you do stuff that you want a feedback on. I can't understand that from a single sentence in PR description and another sentence in the comment.

My questions are: Who is the orchestrator (doing the invocation to the individual functions)? How do you get the workflow structure? Is this DAG mode platform-specific? Maybe we can just add the part where DAGMode reads recorded DAGs instead of creating the new platform.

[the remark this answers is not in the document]
tobystocki: My bad I should have provided more information: 
- We have extended the dirigent cluster manager to be able to orchestrate workflows and want to benchmark it in order to further optimize it. So orchestration is done within dirigent and the workflows are to be provided when registering them in the system. From outside the system registering workflows is different to registering a single function but invoking either one looks the same to the user.
- The system is designed to work with dandelion as the underlying runtime. The workflows we want to run are simple data queries where the data operators are the nodes in the DAG.
- The workflow description is exported as a file that is read and the content is given to dirigent in the registration. The structure is not platform specific (though only works on dirigent) but as already said the orchestration inside dirigent only supports the dandelion platform.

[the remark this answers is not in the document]
cvetkovic: > Maybe we can just add the part where DAGMode reads recorded DAGs instead of creating the new platform.

@leokondrashov: This was my thought as well. That's why we asked you for feedback if you can provide your opinion on integration with the `DAGMode`.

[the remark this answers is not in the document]
leokondrashov: Okay, that makes sense. Since it affects the deployment, not the invocations, I'd say the DAGMode is not very suitable. However, merging it with platform is not a great idea either. Can we reuse the yamlselector for such deployment options? Right now, it is used only for knative deployment to choose from container and firecracker, so it is useless in dirigent case. But we can have dirigent-specific options for deployment. This way, we would be able to remove dandelion from the platform settings as well. This change would require renaming the option since it would mean deployment type.

BTW, please add all the points you wrote in the comments to the PR description and docs. This is useful from the end-user perspective and ours, just to remember what and why we have these features.

pkg/driver/clients/http_client.go
```diff
@@ -85,6 +66,60 @@ func (i *httpInvoker) Invoke(function *common.Function, runtimeSpec *common.Runt
...
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Host = wf.Name // dirigent takes request name from this
+
+	return req
+}
+
+func (i *httpInvoker) Invoke(function *common.Function, runtimeSpec *common.RuntimeSpecification) (bool, *mc.ExecutionRecord) {
+	isDandelion := strings.Contains(strings.ToLower(i.cfg.Platform), "dandelion")
+	isKnative := strings.Contains(strings.ToLower(i.cfg.Platform), "knative")
+	isWorkflow := strings.Contains(strings.ToLower(i.cfg.Platform), "workflow")
```
leokondrashov, line 93: This is done literally on each request, seems quite excessive. Can we do something like it is done in [this](https://github.com/vhive-serverless/invitro/pull/585) PR?

leokondrashov (commented):
Overall comment, can you add tests for the deployment? I don't know how hard it is to test the dirigent and dandelion setup in github actions, but better to have them.

cmd/loader.go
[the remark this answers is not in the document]
cvetkovic: @leokondrashov, @tobystocki: It seems we have come to the point where we have a lot of Dirigent-specific configuration knobs, so I would go ahead and extract the Dirigent-specific knobs into a separate JSON file, similar to `failure.json`.

I would suggest we have the following parameters in the `dirigent_config.json` (don't name it `dirigent.json`, as that conflicts with trace metadata):
- backend (containerd/Firecracker/Dandelion)
- workflow (true/false)
- DirigentControlPlaneIP
- BusyLoopOnSandboxStartup (perhaps used in Knative - @leokondrashov)
- AsyncMode
- AsyncResponseURL
- AsyncWaitToCollectMin

pkg/driver/clients/http_client.go
cvetkovic: Agree.

tobystocki: Fixed in [06f9a23](https://github.com/vhive-serverless/invitro/pull/605/commits/06f9a23db9d6549d0e3bbdaed8292b1326222198) - this is now resolved on initialization.

pkg/driver/clients/invoker.go
```diff
@@ -23,7 +23,7 @@ func CreateInvoker(cfg *config.LoaderConfiguration, announceDoneExe *sync.WaitGr
 		} else {
 			return newHTTPInvoker(cfg)
 		}
-	case "Dirigent-Dandelion":
+	case "Dirigent-Dandelion", "Dirigent-Dandelion-Workflow":
```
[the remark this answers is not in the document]
tobystocki: Resolved in [06f9a23](https://github.com/vhive-serverless/invitro/pull/605/commits/06f9a23db9d6549d0e3bbdaed8292b1326222198)

cmd/loader.go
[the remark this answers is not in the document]
tobystocki: Alright I made the following changes ([332bef6](https://github.com/vhive-serverless/invitro/commit/332bef6f7fcee7b4b609522f4f10534e5018335c)):

- The config params from above were moved into a separate dirigent config file.
- Only one dirigent platform now. If you choose dirigent as the platform you need to provide a dirigent config path from which the dirigent config is loaded (some examples are under `cmd/dirigent`). Loading the dirigent config is currently done when creating a `config.Configuration` in the `runTraceMode` and `runRPSMode` functions.
- Everything workflow related is now in the dirigent config and the workflow config (which is specific to each workflow).

pkg/driver/clients/dandelion_interface.go
```diff
@@ -87,18 +90,95 @@ func composeBusyLoopBody(functionName, image string, runtime, iterations int) *b
...
+			if itm == "" {
+				data = []byte{}
+				ident = "empty"
+			} else {
+				var err error
+				if strings.HasPrefix(itm, "%path=") {
+					logrus.Tracef("Detected local path for item %d in set %d (%s)", itmIdx, setIdx, itm)
+					data, err = os.ReadFile(itm[6:])
+					if err != nil {
+						logrus.Fatalf("Failed to read file '%s': %v", itm[6:], err)
```
cvetkovic, line 120: Might not be that fatal, since it's a per-request thing. Error instead?

Also, do you really want to read the file on the critical path, especially if it's the same for all the requests.

pkg/driver/clients/http_client.go
```diff
@@ -26,14 +26,27 @@ type FunctionResponse struct {
...
+
 	return &httpInvoker{
-		client: CreateHTTPClient(cfg.GRPCFunctionTimeoutSeconds, cfg.InvokeProtocol),
-		cfg:    cfg,
+		client:      CreateHTTPClient(lcfg.GRPCFunctionTimeoutSeconds, lcfg.InvokeProtocol),
+		loaderCfg:   lcfg,
+		dirigentCfg: dcfg,
+
+		isKnative:   strings.Contains(strings.ToLower(lcfg.Platform), "knative"),
+		isDandelion: strings.Contains(strings.ToLower(dcfg.Backend), "dandelion"),
```
cvetkovic, line 48: Can you extract all these platform strings into global constants in the whole repo?

pkg/driver/deployment/deployer.go
```diff
@@ -14,8 +14,8 @@ func CreateDeployer(cfg *config.Configuration) FunctionDeployer {
 	switch cfg.LoaderConfiguration.Platform {
 	case "AWSLambda":
 		return newAWSLambdaDeployer()
-	case "Dirigent", "Dirigent-Dandelion":
-		return newDirigentDeployer()
+	case "Dirigent":
+		return newDirigentDeployer(cfg.DirigentConfiguration.Workflow)
```
cvetkovic, line 18: This should be put into `dirigentDeploymentConfiguration` data structure.

pkg/driver/clients/invoker.go
```diff
@@ -13,24 +13,25 @@ type Invoker interface {
...
 	case "AWSLambda":
 		return newAWSLambdaInvoker(announceDoneExe)
 	case "Dirigent":
-		if cfg.InvokeProtocol == "grpc" {
-			return newGRPCInvoker(cfg, ExecutorRPC{})
-		} else {
+		if cfg.DirigentConfiguration == nil {
+			logrus.Fatal("Failed to create invoker: dirigent configuration is required for platform 'dirigent'")
+		}
+		if cfg.DirigentConfiguration.Backend == "Dandelion" || cfg.LoaderConfiguration.InvokeProtocol != "grpc" {
```
cvetkovic, line 24: Should be checking for lowercase?

cvetkovic (changes requested):

pkg/config/parser.go
```diff
@@ -47,23 +47,13 @@ type LoaderConfiguration struct {
...
 	RpsTarget                   float64 `json:"RpsTarget"`
 	RpsColdStartRatioPercentage float64 `json:"RpsColdStartRatioPercentage"`
 	RpsCooldownSeconds          int     `json:"RpsCooldownSeconds"`
 	RpsImage                    string  `json:"RpsImage"`
 	RpsRuntimeMs                int     `json:"RpsRuntimeMs"`
 	RpsMemoryMB                 int     `json:"RpsMemoryMB"`
-	RpsRequestedGpu             int     `json:"RpsRequestedGpu"`
 	RpsIterationMultiplier      int     `json:"RpsIterationMultiplier"`
-	RpsDataSizeMB               float64 `json:"RpsDataSizeMB"`
-	RpsFile                     string  `json:"RpsFile"`
```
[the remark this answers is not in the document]
leokondrashov: 1. I don't see any docs on RpsFile parameter at all. So, it's hard to tell whether this one is relevant.
2. I'd say that we should leave options that are reasonably extendable to a general platform (iteration multiplier, memory, duration, payload size) in the common config. The rest should go to Dirigent-specific, GPU, for example. If someone implements one of the latter parameters for knative/ow/aws, then we pull it from dirigent-specific parameters into common.

cmd/config_dirigent_dandelion_rps.json
```diff
@@ -1,25 +1,18 @@
...
-  "AsyncWaitToCollectMin": 1,
-
   "RpsTarget": 1,
   "RpsColdStartRatioPercentage": 0,
   "RpsCooldownSeconds": 10,
   "RpsImage": "empty",
   "RpsRuntimeMs": 10,
   "RpsMemoryMB": 2048,
   "RpsIterationMultiplier": 80,
-  "RpsDataSizeMB": 0.00,
```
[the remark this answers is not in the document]
tobystocki: This was introduced in the PR for dirigent GPUs and is only used there. That's why I put it into the dirigent config. Should this be a common config parameter?

pkg/driver/clients/dandelion_interface.go
tobystocki: For workflows I'm precomputing the requests (for each of the workflows, if multiple) when registering them. So this is not on the critical path.

pkg/driver/clients/http_client.go
tobystocki: I could but does that make sense. After all they are only used in the http client.

pkg/config/parser.go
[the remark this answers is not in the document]
tobystocki: Those three parameters were introduced in the latest PR on gpu support for dirigent. Since they are dirigent specific for now I'll leave them in the dirigent only config. I also added the params to the docs.

pkg/driver/deployment/deployer.go
tobystocki: I don't understand - the `Workflow` parameter is part of the dirigent config?

pkg/driver/clients/invoker.go
tobystocki: Fixed in [3a3cf70](https://github.com/vhive-serverless/invitro/pull/605/commits/3a3cf705e76b575ae7414cab6bf31d80453313ff)

docs/configuration.md
```diff
@@ -1,43 +1,38 @@
...
+|------------------------------|-----------|---------------------------------------------------------------------|---------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Seed                         | int64     | any                                                                 | 42                  | Seed for specification generator (for reproducibility)                                                                                                                                                                                   |
+| Platform                     | string    | Knative, OpenWhisk, AWSLambda, Dirigent, Dirigent-Dandelion         | Knative             | The serverless platform the functions will be executed on                                                                                                                                                                                |
```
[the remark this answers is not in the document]
tobystocki: Fixed in [3a3cf70](https://github.com/vhive-serverless/invitro/pull/605/commits/3a3cf705e76b575ae7414cab6bf31d80453313ff)

cmd/loader.go
```diff
@@ -151,7 +151,9 @@ func parseYAMLSpecification(cfg *config.LoaderConfiguration) string {
 	case "firecracker":
 		return "workloads/firecracker/trace_func_go.yaml"
 	default:
-		if cfg.Platform != "Dirigent" && cfg.Platform != "Dirigent-Dandelion" {
+		platform := strings.ToLower(cfg.Platform)
+		validPlatforms := []string{"dirigent", "dirigent-dandelion", "dirigent-dandelion-workflow"}
```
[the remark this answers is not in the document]
tobystocki: Fixed in [3a3cf70](https://github.com/vhive-serverless/invitro/pull/605/commits/3a3cf705e76b575ae7414cab6bf31d80453313ff)

pkg/driver/clients/http_client.go
cvetkovic: "knative", "dandelion", "dirigent" string constants are used throughout the codebase. I thought to extract that.

pkg/driver/deployment/deployer.go
cvetkovic: You should do it like this https://github.com/vhive-serverless/invitro/blob/49628dc5f77e15c35fbe76ff38d003ca97448016/pkg/driver/deployment/dirigent.go#L25.

tobystocki: Resolved in [30dd4da](https://github.com/vhive-serverless/invitro/pull/605/commits/30dd4daf6e2f661355f0dc22a9a0cabf5cee1af5)

pkg/driver/clients/http_client.go
tobystocki: Resolved in [30dd4da](https://github.com/vhive-serverless/invitro/pull/605/commits/30dd4daf6e2f661355f0dc22a9a0cabf5cee1af5)

cmd/config_dirigent_dandelion_rps.json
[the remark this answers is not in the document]
tobystocki: Will remain a dirigent specific one since it is only used in the case where RpsRequestedGpu is >0 and RpsFile is empty (both are dirigent specific parameters and thus in the dirigent config)

cmd/loader.go
```diff
@@ -151,7 +149,7 @@ func parseYAMLSpecification(cfg *config.LoaderConfiguration) string {
 	case "firecracker":
 		return "workloads/firecracker/trace_func_go.yaml"
 	default:
-		if cfg.Platform != "Dirigent" && cfg.Platform != "Dirigent-Dandelion" {
+		if cfg.Platform == common.PlatformDirigent {
 			log.Fatal("Invalid 'YAMLSelector' parameter.")
 		}
```
cvetkovic, line 154: Should be `!=`?

cvetkovic (commented):
LGTM! Just a small bugfix is needed. If @leokondrashov agrees, we can merge.

cmd/loader.go
tobystocki: Yes, fixed in [15ee204](https://github.com/vhive-serverless/invitro/pull/605/commits/15ee204c5b1b2c4e26fe7a6e8521bd771e0cad03)

docs/configuration.md
```diff
@@ -1,43 +1,38 @@
...
+| RpsTarget                    | int       | >= 0                                                                | 0                   | Number of requests per second to issue                                                                                                                                                                                                   | 
+| RpsColdStartRatioPercentage  | int       | >= 0 && <= 100                                                      | 0                   | Percentage of cold starts out of specified RPS                                                                                                                                                                                           | 
+| RpsCooldownSeconds [^6]      | int       | > 0                                                                 | 0                   | The time it takes for the autoscaler to downscale function (higher for higher RPS)                                                                                                                                                       |
```
leokondrashov, line 13: Footnote shows broken in rich view on github; shows as `[^6]`, no link; all the rest are fine.

docs/configuration.md
```diff
@@ -1,43 +1,38 @@
...
+| ExperimentDuration           | int       | > 0                                                                 | 1                   | Experiment duration in minutes of trace to execute excluding warmup                                                                                                                                                                      |
+| WarmupDuration               | int       | > 0                                                                 | 0                   | Warmup duration in minutes(disabled if zero)                                                                                                                                                                                             |
+| PrepullMode                  | string    | all_sync, all_async, one_sync, one_async, none                      | none                | Prepull image before starting experiments sync or async                                                                                                                                                                                  |
```
leokondrashov, line 25: Dirigent-specific, I think should be moved to Dirigent config

docs/configuration.md
```diff
@@ -1,43 +1,38 @@
...
+| RpsColdStartRatioPercentage  | int       | >= 0 && <= 100                                                      | 0                   | Percentage of cold starts out of specified RPS                                                                                                                                                                                           | 
+| RpsCooldownSeconds [^6]      | int       | > 0                                                                 | 0                   | The time it takes for the autoscaler to downscale function (higher for higher RPS)                                                                                                                                                       |
+| RpsImage                     | string    | N/A                                                                 | N/A                 | Function image to use for RPS experiments                                                                                                                                                                                                |
```
leokondrashov, line 14: This one is only used for Dirigent; for knative we have image specified in the yaml file.

For this one there are two options: make it completely Dirigent-specific and move to its config or extend it to be applicable in knative (by replacing the image in the yaml as we do with couple of other parameters). Right now I don't see much reason for the latter, TBH.

leokondrashov (commented):
Looks good, please fix the docs and couple of options

docs/configuration.md
tobystocki: Done in [7b6fcb0](https://github.com/vhive-serverless/invitro/pull/605/commits/7b6fcb032c6bf2401fcfcb3cc6e4535094573372)

pkg/generator/rps.go
```diff
@@ -105,7 +105,7 @@ func GenerateColdStartFunctions(experimentDuration int, rpsTarget float64, coold
 	return functions, countResult
 }
 
-func CreateRPSFunctions(cfg *config.LoaderConfiguration, warmFunction common.IATArray, warmFunctionCount []int,
+func CreateRPSFunctions(cfg *config.LoaderConfiguration, dcfg *config.DirigentConfig, warmFunction common.IATArray, warmFunctionCount []int,
```
leokondrashov, line 108: What would happen if we have no Dirigent config set in the options? It seems that ReadDirigentConfig would return nil in that case, so it would be passed here. I don't see anything that would catch that.

tobystocki: You're right, I fixed it in [73b3de6](https://github.com/vhive-serverless/invitro/pull/605/commits/73b3de6624368ce7f161c1b2a803ef6364a16abb)

cvetkovic (dismissed):

cvetkovic (dismissed):

cvetkovic (approved):

cvetkovic (approved):

Issue #586, opened 2025-01-30
High warm start delay


leokondrashov:
## Version

`main` branch

## Steps to reproduce

Run RPS mode with all warm start (`RpsColdStartRatioPercentage` is 0) and `RPS` of 1.

## Expected behavior

Fast warmups after stabilization period (~5ms)

## Observed behavior

Big delay, ~50ms

## Additional notes

It's been reported that this problem diminishes with higher RPS.
A possible cause is the switch to new gRPC invocation semantics using client instead of connection establishment.

leokondrashov:
With a higher invocation rate (e.g., 10RPS), the problem diminishes but doesn't go away completely. It still affects ~10% of invocations.

PR #618, opened 2025-03-16
Change to passthrough resolver

dhairyarungta:
Title: Fix high warm start delay by using passthrough resolver

This PR addresses issue #586 where we're experiencing high warm start delays (~50ms) with gRPC connections, particularly at low RPS rates (1 RPS).

When using `grpc.NewClient()` with just the endpoint, it defaults to DNS resolution which adds significant latency to the first connection. This behavior differs from the older `grpc.Dial()` method which defaulted to passthrough resolution.

Solution:
The fix is to explicitly use the passthrough resolver by prefixing the endpoint with "passthrough:///". This bypasses DNS resolution and sends the endpoint string verbatim to the dialer, which significantly reduces connection establishment time.

This change is based on the discussion in grpc/grpc-go#1786 which explains the differences between Dial and NewClient behaviors regarding resolvers.

### DIFF
--- a/pkg/driver/clients/grpc_client.go
+++ b/pkg/driver/clients/grpc_client.go
@@ -146,7 +146,7 @@ func (i *grpcInvoker) Invoke(function *common.Function, runtimeSpec *common.Runt
 
 	grpcStart := time.Now()
 
-	conn, err := grpc.NewClient(function.Endpoint, dialOptions...)
+	conn, err := grpc.NewClient("passthrough:///" + function.Endpoint, dialOptions...)
 	if err != nil {
 		logrus.Debugf("Failed to establish a gRPC connection - %v\n", err)
 


leokondrashov:
@cvetkovic this is the fix we came up with for the grpc delays. It works well with knative setup, warm starts are ~5ms again. Can you please check with Dirigent? My concern is the format of the endpoints might be different, so simple concatenation might break the invoker in some cases.

cvetkovic (approved):

cvetkovic:
> @cvetkovic this is the fix we came up with for the grpc delays. It works well with knative setup, warm starts are ~5ms again. Can you please check with Dirigent? My concern is the format of the endpoints might be different, so simple concatenation might break the invoker in some cases.

Works on Dirigent!

PR #621, opened 2025-03-25
YAMLSelector bugfix for Knative RPS

cvetkovic:
Removing redundant fields that added consistency problems.

### DIFF
--- a/cmd/loader.go
+++ b/cmd/loader.go
@@ -207,7 +207,6 @@ func runTraceMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIA
 		TraceGranularity: parseTraceGranularity(cfg),
 		TraceDuration:    durationToParse,
 
-		YAMLPath: yamlPath,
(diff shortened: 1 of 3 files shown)


leokondrashov (approved):
Looks fine to me right now.

However, I would argue that we should make it possible to create Azure trace parser and rps generator without passing knative-specific and/or dirigent-specific arguments. At least, we should extract parsing the trace/generating the function inv data from the filling all of the necessary fields in the Function structure.

cvetkovic:
I wanted to remove the `YAMLSelector` from the Function specification, but vSwarm traces rely on this since functions do not share the YAML file, therefore I left it.

leokondrashov:
My point was that we can generate functions with all invocations first and then fill the rest of the fields in a separate method. That would let us generate the invocations without requiring any system-specific parameters/arguments since it is system-agnostic. The function structure doesn't need to be changed; we can leave the fields as nils during the invocation generation phase.

PR #622, opened 2025-03-31
Fix -> Invoke both trace-func and vSwarm functions in the same run

aryans1204:
## Summary

This is a fix for allowing Loader to invoke both trace-func and vSwarm functions in the same run.

The `Invoker` interface is moved to pkg `common` and the Invoker interface is removed from the driver config and instead moved to the `common.Function`. Similarly, the `record.go` file containing the `ExecutionRecord` is moved to `common`, to allow for the ExecutionRecord to be used in the Invoke interface call, which would not be possible otherwise since the `metric` package cant be imported in `common` due to cyclic dependency. Most other files are changed to reflect this change of package from `metric` to `common`. The `trace_driver.go` is changed to generate clients for each parsed function based on `Execute` or `SayHello` and the `invokeFunction` is changed to invoke the function specific client. No significant changes are made to the `Invoker` client, and no chnages are made to the deployment stages. For the trace parsing, under `mapper_trace_parser.go` the YAML file for trace-func is embedded if the chosen function is trace-func.

* 

* 


pkg/driver/trace_driver.go
```diff
@@ -501,6 +504,8 @@ func (d *Driver) RunExperiment() {
 
 	go failure.ScheduleFailure(d.Configuration.LoaderConfiguration.Platform, d.Configuration.FailureConfiguration)
 
+	//Generate clients
```
cvetkovic, line 507: Space after `//`.

pkg/driver/trace_driver_test.go
```diff
@@ -530,35 +535,6 @@ func TestDriverCompletely(t *testing.T) {
...
-			traceGranularity:    common.SecondGranularity,
-			withWarmup:          true,
-			expectedInvocations: 120,
-		},
-		{
-			testName:              "without_warmup_sleep_1min_then_invoke",
-			experimentDurationMin: 2,
-			invocationStats:       []int{0, 5},
-			expectedInvocations:   5,
-		},
```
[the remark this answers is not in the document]
aryans1204: I deleted the three tests because the trace driver unit test was taking over 10 minutes for completion. Alternatively, I can increase the timeout parameter for the test in the unit-tests.yaml.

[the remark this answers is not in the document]
aryans1204: I added the tests back and increased the timeout limit in the unit tests

pkg/driver/trace_driver.go
```diff
@@ -61,16 +60,21 @@ func NewDriver(driverConfig *config.Configuration) *Driver {
...
 	}
 
-	d.Invoker = clients.CreateInvoker(driverConfig, &d.allFunctionsInvoked, &d.readOpenWhiskMetadata)
-
 	return d
 }
 
+func (d *Driver) generateInvokers() {
+	for _, function := range d.Configuration.Functions {
+		trace_func := strings.Contains(function.Name, "trace-func")
```
leokondrashov, line 73: I wouldn't rely on function name here. In RPS mode, we have functions named differently even for the trace functions. BTW, we would need to think how to pass this mapping in that case as well.

pkg/metric/cluster_metrics.go
leokondrashov, line 1: Now you duplicate it both in common and metric package. Please leave them only here.

PR #740, opened 2026-03-03
Refactor Loader.go, in preparation for adding new trace type.

16fb:
## Summary
Refactored Loader.
Ensure more clear separation of function parsing/generation, dirigent reading, driver running. 

Added traceInputType variable classification
Combined runRPS() and runTrace() into singular runMode()

Driver should use []*common.function directly, extracted function generation out of driver.
Driver should not handle function read/write to file, extract out of driver.

Dirigent metadata adding extracted out of function generation, takes a dedicated section of runMode()
RPS function generation rewritten to extract dirigent metadata adding out.

N/A

trace_driver_test had to be updated to use new function specification generation method.
driver.GenerateSpecification() => GenerateAzure2019Specification()

### DIFF
--- a/cmd/loader.go
+++ b/cmd/loader.go
@@ -144,6 +144,7 @@ func parseIATDistribution(cfg *config.LoaderConfiguration) (common.IatDistributi
 	return common.Exponential, false
 }
 
+// Return YAML for "container" or "firecracker microVM"
 func parseYAMLSpecification(cfg *config.LoaderConfiguration) string {
 	switch cfg.YAMLSelector {
 	case "container":
@@ -270,6 +271,9 @@ func runRPSMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATs
 }
 
 func runMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATsToFile bool) {
+	//
+	// Determine type of input.
+	//
 	var traceInputType string
 	if cfg.TracePath == "RPS" {
 		traceInputType = "RPS"
@@ -278,9 +282,11 @@ func runMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATsToF
 	} else { // Reduant, for future input types.
 		log.Fatal("Unsupported Trace Input Type", traceInputType)
 	}
+	log.Info("Detected Trace Input Type", traceInputType)
+
 
 	//
-	// Generate common.Functions
+	// Generate common.Functions + FunctionSpecification (Function's deployment and invocation info)
 	//
 	var functions []*common.Function
 	switch traceInputType {
@@ -288,8 +294,23 @@ func runMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATsToF
 		functions = RPSGenerateFunctions(cfg)
 	case "Azure2019":
 		functions = Azure2019GenerateFunctions(cfg)
-	default:
-		log.Fatal("Unsupported Trace Input Type:", traceInputType)
+	}
+
+	//
+	// Handle Dirgient (Container orchestrator)
+	//
+	dirigentConfig := config.ReadDirigentConfig(cfg)
+	switch traceInputType {
+	case "RPS":
+		// Handle dirigent metadata, handle warm and cold function differently.
+		functions = generator.AppendDirigentMetadata(cfg, dirigentConfig, functions)
+	case "Azure2019":
+		// Reads Dirigent trace meta-data, and places into *common.Function as property "dirigentMetadata"
+		// Or if Knative yaml is provided, converts it to dirigent configurations. 
+		// Dirigent metadata parsing
+		yamlPath := parseYAMLSpecification(cfg)
+		dirigentMetadataParser := trace.NewDirigentMetadataParser(cfg.TracePath, functions, yamlPath, cfg.Platform)
+		dirigentMetadataParser.Parse()
 	}
 
 	log.Infof("Traces contain the following %d functions:\n", len(functions))
@@ -299,9 +320,8 @@ func runMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATsToF
 
 	driver.ReadOrWriteSpecificationToFile(functions, writeIATsToFile, readIATFromFile)
 
-	experimentDuration := determineDurationToParse(cfg.ExperimentDuration, cfg.WarmupDuration)
-	dirigentConfig := config.ReadDirigentConfig(cfg)
 
+	experimentDuration := determineDurationToParse(cfg.ExperimentDuration, cfg.WarmupDuration)
 	experimentDriver := driver.NewDriver(&config.Configuration{
 		LoaderConfiguration:   cfg,
 		TraceDuration:         experimentDuration,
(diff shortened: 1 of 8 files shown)


pkg/driver/trace_driver.go
```diff
@@ -433,62 +433,69 @@ func (d *Driver) internalRun() {
 	log.Infof("Failure rate: \t\t\t%.2f%%", float64(statFailed)*100.0/float64(statSuccess+statFailed))
 }
 
-func (d *Driver) GenerateSpecification() {
+// Generates IATs and runtime specifications for Azure2019 trace type, returns completed `functions` with Specification filled.
+func GenerateAzure2019Specification(functions []*common.Function, loaderCfg *config.LoaderConfiguration, IATDistribution common.IatDistribution, shiftIAT bool, traceGranularity common.TraceGranularity) []*common.Function {
```
leokondrashov: you pass and return the same thing (functions), you also update it inplace, so returning it is redundant

pkg/generator/rps.go
```diff
@@ -179,6 +155,52 @@ func CreateRPSFunctions(cfg *config.LoaderConfiguration, dcfg *config.DirigentCo
 	return result
 }
 
+// TODO consider pointer receiver for []*functions
+
+// Attaches 2 possible DirigentMetadata property, depending if function is cold or warm (based on function name) for RPS function.
+func AppendDirigentMetadata(cfg *config.LoaderConfiguration, dcfg *config.DirigentConfig, functions []*common.Function) []*common.Function {
```
leokondrashov: same here, no need to pass and return the same thing

pkg/driver/trace_driver.go
leokondrashov: this might be more suitable to move to generator module

cmd/loader.go
```diff
@@ -105,11 +105,12 @@ func main() {
...
-		runRPSMode(&cfg, *iatFromFile, *iatGeneration)
-	} else {
-		runTraceMode(&cfg, *iatFromFile, *iatGeneration)
-	}
+	runMode(&cfg, *iatFromFile, *iatGeneration)
+	// if cfg.TracePath == "RPS" {
+	// 	runRPSMode(&cfg, *iatFromFile, *iatGeneration)
+	// } else {
+	// 	runTraceMode(&cfg, *iatFromFile, *iatGeneration)
+	// }
```
leokondrashov: remove unnecessary comments

cmd/loader.go
```diff
@@ -171,61 +173,78 @@ func parseTraceGranularity(cfg *config.LoaderConfiguration) common.TraceGranular
...
-
-		TestMode: false,
-
-		Functions: functions,
+	experimentDuration := determineDurationToParse(cfg.ExperimentDuration, cfg.WarmupDuration)
+	experimentDriver := driver.NewDriver(&config.Configuration{
+		LoaderConfiguration:   cfg,
+		TraceDuration:         experimentDuration,
+		DirigentConfiguration: dirigentConfig,
+		Functions:             functions,
```
leokondrashov: the experiment driver misses quite a few things in creation. If this becomes redundant with the new generation, they should be removed from configuration structure, otherwise need to return them here

16fb: Some of the struct fields are only used for function specification generation (IATDistribution and ShiftIAT).

While these fields are referenced in driver_test functions, i will update these test functions to derive their values directly.

16fb:
Previous 2 commits added your pointers to the PR
The 3 commits before that, fixed the PR automated test popups

cmd/loader.go
```diff
@@ -171,61 +168,83 @@ func parseTraceGranularity(cfg *config.LoaderConfiguration) common.TraceGranular
...
+	//
+	dirigentConfig := config.ReadDirigentConfig(cfg)
+	switch traceInputType {
+	case "RPS":
+		// Handle dirigent metadata, handle warm and cold function differently.
+		generator.AppendDirigentMetadata(functions, cfg, dirigentConfig)
+	case "Azure2019", "vSwarm":
+		// Reads Dirigent trace meta-data, and places into *common.Function as property "dirigentMetadata"
+		// Or if Knative yaml is provided, converts it to dirigent configurations. 
+		// Dirigent metadata parsing
```
leokondrashov: please go over the comments you added and prune them. Some of them are really not needed like this "Dirigent metadata parsing" are now redundant.

cmd/loader.go
```diff
@@ -171,61 +168,83 @@ func parseTraceGranularity(cfg *config.LoaderConfiguration) common.TraceGranular
...
-	iatType, shiftIAT := parseIATDistribution(cfg)
+	driver.ReadOrWriteSpecificationToFile(functions, writeIATsToFile, readIATFromFile)
+
 
+	experimentDuration := determineDurationToParse(cfg.ExperimentDuration, cfg.WarmupDuration)
 	experimentDriver := driver.NewDriver(&config.Configuration{
-		LoaderConfiguration:  cfg,
+		LoaderConfiguration:   cfg,
 		FailureConfiguration: config.ReadFailureConfiguration(*failurePath),
+		DirigentConfiguration: dirigentConfig,
```
leokondrashov, line 225: did you run through gofmt? it seems weird that cfg got moved by one space but failure confif wasn't

cmd/loader.go
```diff
@@ -171,61 +168,83 @@ func parseTraceGranularity(cfg *config.LoaderConfiguration) common.TraceGranular
 	return common.MinuteGranularity
 }
 
-func runTraceMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATsToFile bool) {
-	durationToParse := determineDurationToParse(cfg.ExperimentDuration, cfg.WarmupDuration)
-	yamlPath := parseYAMLSpecification(cfg)
-	var functions []*common.Function
-	var traceParser trace.Parser
+func runMode(cfg *config.LoaderConfiguration, readIATFromFile bool, writeIATsToFile bool) {
```
leokondrashov: I would call it just run at this point

pkg/generator/rps.go
```diff
@@ -105,33 +108,20 @@ func GenerateColdStartFunctions(experimentDuration int, rpsTarget float64, coold
 	return functions, countResult
 }
 
-func CreateRPSFunctions(cfg *config.LoaderConfiguration, dcfg *config.DirigentConfig, warmFunction common.IATArray, warmFunctionCount []int,
+// Generates []*common.Function based on intended warm/cold functionality, does not generate dirigent metadata.
```
leokondrashov: I would remove the part of the comment about what it doesn't do

pkg/generator/rps.go
```diff
@@ -179,6 +155,50 @@ func CreateRPSFunctions(cfg *config.LoaderConfiguration, dcfg *config.DirigentCo
 	return result
 }
 
+// TODO consider pointer receiver for []*functions
```
leokondrashov: I think this one is redundant

leokondrashov (commented):
Quite good now, please fix minor comments and ready to merge

cmd/loader.go
16fb: Turns out i did not have gofmt running automatically. I have enabled it to run on file save now.

leokondrashov (approved):
Looks good. Need to rebase on the new main, shouldn't be a problem since the update was only with the k8s setup, without touching the loader code

16fb:
Yeap, have rebased on new main

PR #751, opened 2026-06-02
Allow Sampler to work with Azure2021 Function Traces

16fb:
## Summary

Update Sampler to use Azure2021 function traces.

Added `preprocessor2021`, transforms azure2021 trace to format usable by `sample`.
Utilize `sample` to generate samples, generating list of functions to keep.
Added `filter`, which notes down the list of functions to keep, and filters the original azure2021 trace, 
keeping the functions in the generated list.

No new python dependencies not already in `requirements.txt` 

none (N/A)

### DIFF
--- a/docs/sampler.md
+++ b/docs/sampler.md
@@ -187,7 +187,7 @@ The differences stems from the differences in trace:
 
 The preprocessing follows 3 steps:
 1. Filter for invocations within time interval (according to the user-defined start and end time)
-2. Remove functions with invocation rate of 0ms duration above threshold rate. (user-defined threshold)
+2. Remove functions with invocation rate of 0ms duration above threshold rate. Default threshold of 50%. (user-defined threshold)
 3. Transform the trace to azure2019 format. 
 
 While `inv_df` and `dur_df` can be calculated, `mem_df` is set to a static value of 200.
--- a/docs/sampler.md
+++ b/docs/sampler.md
@@ -173,7 +173,7 @@ in [`plotTimeline/analysis.ipynb`](/tools/plotTimeline/analysis.ipynb)
 ### Overview
 
 The sampling process of Azure2021 trace follows 3 steps: 
-1. Pre-process the original trace (`preprocess2021`) -> obtain clean trace.
+1. Pre-process the original trace (`preprocess2021`) -> obtain clean trace in azure2019 format.
 2. Run the sampler on cleaned trace (`sampler`) -> obtain down-sampled trace, which contains a list of functions to keep in azure2019 format.
 3. Filter the original trace for functions in list (`filter`) -> obtain down-sampled trace in azure2021 format (per-invocation format).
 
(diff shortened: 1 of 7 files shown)


docs/sampler.md
```diff
@@ -168,3 +168,48 @@ Tools that can be used to generate the timeline of a trace are available in the
...
+
+The differences stems from the differences in trace:
+- azure2021 trace is invocation per row, while azure2019 is a function per row.
+- azure2021 trace does not have memory data.
+
+### Preprocess2021
+
+The preprocessing follows 3 steps:
+1. Filter for invocations within time interval (according to the user-defined start and end time)
+2. Remove functions with invocation rate of 0ms duration above threshold rate. (user-defined threshold)
```
leokondrashov: please put the default value that you use for reference here

docs/sampler.md
```diff
@@ -168,3 +168,48 @@ Tools that can be used to generate the timeline of a trace are available in the
...
+3. Transform the trace to azure2019 format. 
+
+While `inv_df` and `dur_df` can be calculated, `mem_df` is set to a static value of 200.
```
leokondrashov, line 193: is 200 the same as default in the loader?

sampler/tests/filter_test.py
```diff
@@ -0,0 +1,96 @@
...
+import logging as log
+
+from sampler.filter import filter_azure2021
+import os
+
+def create_original_df_file(tmp_path):
+
+    og_df = pd.DataFrame(
+        {
+            "app":           ["aa", "ab", "ac", "ac", "ac"],
```
leokondrashov, line 54: please add the test for same app hash different functions. that is possible in the 2019 trace, so if it repeats in the 2021 trace, we better test for it.

sampler/filter.py
```diff
@@ -0,0 +1,76 @@
...
+
+import os
+from glob import glob
+
+from sampler.preprocess2021 import filter_within_time_interval
+
+def filter_azure2021(orig_trace_dir: str, sampled_trace_dir: str, out_dir: str, start_time: str, duration: int):
+
+    # Read original trace
+    trace_file = glob(f"{orig_trace_dir}/AzureFunctionsInvocationTraceForTwoWeeksJan2021.txt")
```
leokondrashov: can you set a specific file here? it's not as critical, but might be useful to sample further the already sampled trace (which would have a different name at that point). this doesn't require a change in logic, just a config change to accept the specific trace file

sampler/preprocess2021.py
```diff
@@ -0,0 +1,244 @@
+#  MIT License
+#
+#  Copyright (c) 2023 EASL and the vHive community
```
leokondrashov: please update the copyrights to the new date and "HySCALE and vHive community"

leokondrashov (commented):
Looks good, see minor comments below

docs/sampler.md
16fb: No. The default for loader was 150 (I eye-balled the value). I will instead match loader to follow 200MB as well.

sampler/preprocess2021.py
16fb: Do i update all copyrights in python, or just for the new files i have created?

sampler/tests/filter_test.py
```diff
@@ -0,0 +1,96 @@
+#  MIT License
+#
+#  Copyright (c) 2023 EASL and the vHive community
```
leokondrashov: please update the licence

sampler/preprocess2021.py
leokondrashov: only in the new ones

sampler/filter.py
16fb: Sure, letting it be modular is good regardless.

I was thinking that sampling further a sampled trace, would be something done in the `Sampler.py`. Then the final trace would be used in `filter.py`.

Regardless, a user might want to filter from a file that is not the original large .txt, using some other metric. I'll add it in as a optional parameter.

sampler/preprocess2021.py
```diff
@@ -0,0 +1,244 @@
...
+from tqdm import tqdm
+from typing import Tuple
+import math
+
+import os
+
+def preprocess_file(trace_dir: str, start_time: str, duration: str, output_dir: str, zero_ms_threshold_percent: str) -> pd.DataFrame:
+    
+    # Read CSV
+    trace_file = glob(f"{trace_dir}/AzureFunctionsInvocationTraceForTwoWeeksJan2021.txt")
```
leokondrashov: can you do the same as in filter for the arbitrary file name here? i don't want to make it so that only the specific file name could be preprocessed

sampler/__main__.py
```diff
@@ -234,6 +247,104 @@ def main():
...
+        metavar='duration',
+        help='Duration in minutes of the excerpt extracted from the postprocessed trace'
+    )
+
+    filter2021_parser.add_argument(
+        '-tf',
+        '--trace_file',
+        required=False,
+        metavar='trace_filename',
+        help='Filename of the original trace, if using non-default original trace other than `AzureFunctionsInvocationTraceForTwoWeeksJan2021.txt`'
```
leokondrashov: i think that could be combined with the --trace flag. we need to point to a single file either way, so no need to break it down to two flags to do that.

leokondrashov (approved):
lgtm

PR #763, opened 2026-06-13
Bump python version to 3.12

16fb:
## Summary

Bump Python version from 3.9 to 3.12, bumping modules as well. 

Each module in `requirements.txt` is pegged to a newer package version (default latest).
Python version in GitHub workflow files bumped from **3.9** to **3.12**.
Only Sampler failed tests, updated code to fix deprecated functions.

(N/A)

Other python source code may fail silently without alerting tests.


16fb:
Sampler errors were caused by a bump in pandas version (~1.3.5 to 3.0.3):

- LossySetItemError 

- Deprecation of .append(). 
  - [Thread](https://github.com/pandas-dev/pandas/issues/35407)
  - [Suggested fix](https://pandas.pydata.org/pandas-docs/version/1.5/whatsnew/v1.4.0.html#deprecated-dataframe-append-and-series-append)

16fb:
**My concerns**:

* I unsure if the requirements.txt should be `==`.

* If so, how do i determine what is the latest version of the modules i should use? Do i check what module version the runner uses?

* The actions `actions/setup-python@v5` and `actions/cache@v4` runs on NodeJS 20. Will be deprecated soon. Should i be updating these actions to `setup-python@v6` and `cache@v5` in this PR as well? Or should it be in a separate PR?

leokondrashov (commented):
I thought we agreed to bump python to the version available in Ubuntu 24.04 (python3.12) to make it consistent with the setup we support.

FYI, the knative e2e test is flaky, the rerun should help.

16fb:
> I thought we agreed to bump python to the version available in Ubuntu 24.04 (python3.12) to make it consistent with the setup we support.
> 
> FYI, the knative e2e test is flaky, the rerun should help.

You're right, i now realised why i left this PR as a draft. Will make the changes

leokondrashov (approved):

PR #765, opened 2026-06-25
Bump python 3.12 that were missed

16fb:
## Summary
Bump action's python version to 3.12
Is a continuation of [previous PR](https://github.com/vhive-serverless/invitro/pull/763), of which some files were missed. 

Bump yaml action files of `actions/setup-python@v6` to 3.12

none (N/A)

none (N/A)

### DIFF
--- a/.github/workflows/azure_integration_test.yaml
+++ b/.github/workflows/azure_integration_test.yaml
@@ -56,10 +56,10 @@ jobs:
             ./go.mod
             ./go.sum
 
-      - name: Set up Python 3.10
+      - name: Set up Python 3.12
         uses: actions/setup-python@v6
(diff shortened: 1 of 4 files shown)


16fb:
This PR is on the Invitro GitHub, so the test requiring azure keys can run. 

Specifically the errors seem to relate to `failed to create resource group`. Should I be concerned about this and try to look for a fix? Or is it the azure key expiration issue you mentioned before, and I can ignore for now?

leokondrashov:
for now can ignore the azure tests, those are failing because of the keys are expired. those are not critical, so we would fix it at some point.

leokondrashov (approved):

PR #766, opened 2026-06-26
Sampler - Allow sampling of `huawei2023-Private` trace type

16fb:
## Summary

Transform `huawei2023-Private` trace dataset into `azure2019` format.
The Huawei trace can then be treated as an azure2019 trace for Sampler and Loader. 

Only minute-binning has been implemented.
Updated `sampler` module to take custom resource normalization value in CLI when using `sample`, to account for differing value in different datasets.

None (N/A)

None (N/A)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@ In-Vitro is a set of tools for analyzing the performance of serverless cluster d
 
 Standard sampled traces are available in [data/traces/reference](data/traces/reference/) folder in this repository. The traces are sampled from the Azure Functions production traces using the [sampler](sampler) tool. More details on the sampling process can be found [here](docs/sampler.md#reference-traces).
 
+## Supported traces
+- [Azure 2019](https://github.com/Azure/AzurePublicDataset/blob/master/AzureFunctionsDataset2019.md)
+- [vSwarm Benchmarks](https://github.com/vhive-serverless/vSwarm)
+- [Azure 2021](https://github.com/Azure/AzurePublicDataset/blob/master/AzureFunctionsInvocationTrace2021.md)
+- [Huawei 2023 Private](https://github.com/sir-lab/data-release/blob/main/README_data_release_2023.md)
+
 ## Reference our work
 
 ```
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@ Standard sampled traces are available in [data/traces/reference](data/traces/ref
 
 ## Supported traces
 - [Azure 2019](https://github.com/Azure/AzurePublicDataset/blob/master/AzureFunctionsDataset2019.md)
-- [vSwarm Benchmarks](https://github.com/vhive-serverless/vSwarm)
 - [Azure 2021](https://github.com/Azure/AzurePublicDataset/blob/master/AzureFunctionsInvocationTrace2021.md)
 - [Huawei 2023 Private](https://github.com/sir-lab/data-release/blob/main/README_data_release_2023.md)
 
--- a/docs/sampler.md
+++ b/docs/sampler.md
@@ -206,10 +206,56 @@ and extract the CSV files (default location: `data/azure2021/`).
 
 The following is an example usage of the sampler:
 
-```console
+``` bash
 python -m sampler preprocess2021 -t data/azure2021/original_trace_filename.txt -o data/traces/reference/preprocessed_azure2021 -s 00:01:00 -dur 100 -thresh 50
 
 python -m sampler sample -t data/traces/reference/preprocessed_azure2021 -orig data/traces/reference/preprocessed_azure2021 -o data/traces/reference/sampled_azure2021 -min 20 -st 5 -max 45 -tr 16
 
 python -m sampler filter2021 -t data/azure2021/original_trace_filename.txt -st data/traces/reference/sampled_azure2021/samples/40 -o data/traces/reference/filtered2021 -s 00:01:00 -dur 100
 ```
+
+## Using Hua-Wei-2023 Traces
+### Overview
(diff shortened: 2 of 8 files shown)


docs/sampler.md
```diff
@@ -206,10 +206,57 @@ and extract the CSV files (default location: `data/azure2021/`).
...
 python -m sampler preprocess2021 -t data/azure2021/original_trace_filename.txt -o data/traces/reference/preprocessed_azure2021 -s 00:01:00 -dur 100 -thresh 50
 
 python -m sampler sample -t data/traces/reference/preprocessed_azure2021 -orig data/traces/reference/preprocessed_azure2021 -o data/traces/reference/sampled_azure2021 -min 20 -st 5 -max 45 -tr 16
 
 python -m sampler filter2021 -t data/azure2021/original_trace_filename.txt -st data/traces/reference/sampled_azure2021/samples/40 -o data/traces/reference/filtered2021 -s 00:01:00 -dur 100
-```
\ No newline at end of file
+```
+
+## Using Hua-Wei-2023 Traces
```
leokondrashov:
```suggestion
## Using Huawei-2023 Traces
```

i haven't seen such way of spelling before, please apply elsewhere

sampler/tests/preprocessHuawei_test.py
```diff
@@ -0,0 +1,121 @@
...
+        "2":    [NaN, 10.0, 10.0],   # At least 1 memory allocation
+        "3":    [1.0, 2.0, 3.555],   # Test 3dp precision
+    })
+
+    expected_df = pd.DataFrame({
+        "HashFunction":             [ "0",  "2",    "3"],
+        "HashOwner":                [ "0",  "0",    "0"],
+        "HashApp":                  [ "0",  "2",    "3"],
+        "Average":                  [ 1.5, 10.0,  2.185],
+        "Count":                    [   3,    2,      3],
```
leokondrashov, line 108: minor thing

please check whether azure trace was using weighted distribution based on the number of invocations in the time interval. iirc, "count" here was total number of invocations, not number of minutes with invocations, so it might be accounting for the invocation count inside a minute for distribution calculations (essentially, the values are repeated based on number of invocations in this minute)

leokondrashov (commented):
the code is mostly fine, please work on docs. plus add to the loader docs that we support the Huawei trace now with the reference how to run it. also, if it is not there, add it to the main README.md, this is an important thing that we support new traces, need to put it at the front.

about the value for the resource norm factor, it seems that the original one is wrong (you can check the units, it is incorrect to multiply by the avg inv count), your new one is correct, so don't touch it.

sampler/tests/preprocessHuawei_test.py
16fb: Okay, i might not be 100% understanding the question/clarification. 

How Azure2019 obtained it data is from [here](https://github.com/Azure/AzurePublicDataset/blob/master/AzureFunctionsDataset2019.md#function-execution-duration), particularly notes 5 and 7.

Azure2019 does use weighted distribution, it samples every 30s with number of invocations ("count") per interval (30s). Then the final percentiles are derived from the list of samples.

But this is to generate the final run_df, so having "count" be total number of invocations is ok?

sampler/sample.py
```diff
@@ -42,7 +42,7 @@
 # Based on the following estimation:
 # Resources distribution is multiplicative of duration (in ms, avg 1000ms),
 # invocation count per minute (avg <10/min), and memory (in MB, avg 200MB)
-RES_NORM_FACTOR = 1000 * 10 * 100
+RES_NORM_FACTOR_AZURE2019 = 1000 * 10 * 100
```
16fb, line 45: > about the value for the resource norm factor, it seems that the original one is wrong (you can check the units, it is incorrect to multiply by the avg inv count), your new one is correct, so don't touch it.

I realize that the original one is wrong. But that means i should change this Norm Factor  to 100,000 right (1000*100)?

sampler/tests/preprocessHuawei_test.py
leokondrashov: the point would be that to do as azure did, we need to add the invocation count matrix to make the distribution. i don't think that's too important, so can skip

README.md
```diff
@@ -4,6 +4,12 @@ In-Vitro is a set of tools for analyzing the performance of serverless cluster d
 
 Standard sampled traces are available in [data/traces/reference](data/traces/reference/) folder in this repository. The traces are sampled from the Azure Functions production traces using the [sampler](sampler) tool. More details on the sampling process can be found [here](docs/sampler.md#reference-traces).
 
+## Supported traces
+- [Azure 2019](https://github.com/Azure/AzurePublicDataset/blob/master/AzureFunctionsDataset2019.md)
+- [vSwarm Benchmarks](https://github.com/vhive-serverless/vSwarm)
```
leokondrashov: vswarm is not a separate trace, but a mode of execution of azure2019 trace with real benchmarks instead of synthetic one, so please remove it from the trace list here and from the loader.md.

overall, it should be possible to run it with any other trace types (because they describe the invocation load) with both sythetic (default) and mapped to a real benchmark set. we can add this as a future feature to test and make it work properly.

leokondrashov (commented):
fix small comment on the vswarm part, and see if the change of the distribution weighting makes a big difference. i would be fine without it as well, pointed it out just for the rigor that might not be very important.

sampler/sample.py
leokondrashov: leave it as it is right now, because changing it would require updating the sampled traces we already have. overall, it seems that nobody cares too much about the sampling methods, so this discrepancy is not important.

leokondrashov (approved):
looks good now

PR #772, opened 2026-07-14
Fix workflow yaml indentation bug.

16fb:
## Summary

Fix syntax error in `tools-tests.yaml` and `code-quality.yml` 
Bump `golanglintci action` to v9.3
Update golanci-lint to v2

* Fixed indentation error in `invitro\.github\workflows\tools-tests.yaml` and `invitro\.github\workflows\code-quality.yml` workflow yaml files.
* Bump `golanglintci action` to latest (v9.3) for Node 24 compatibility.
* Update` invitro\.golangci.yml` to golanci-lint v2. (Using migration tool)

none (N/A)

none (N/A)

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -11,27 +11,15 @@ env:
   GO111MODULE: on
 
 jobs:
-  resolve-modules:
-    name: Resolve Modules
-    runs-on: ubuntu-24.04
-    outputs:
-      matrix: ${{ steps.set-matrix.outputs.matrix }}
-    steps:
-      - name: Checkout Sources
-        uses: actions/checkout@v6
-      - id: set-matrix
-        run: ./utils/golangci-lint/resolve-modules.sh
-
   golangci:
     name: GolangCI Lint
-    needs: resolve-modules
     runs-on: ubuntu-24.04
     strategy:
       matrix: { dir: ['cmd', 'pkg', 'tools/plotter'] }
       fail-fast: false
     steps:
       - name: Checkout code into go module directory
-          uses: actions/checkout@v6
+        uses: actions/checkout@v6
     
       - name: Set up Go version in go.mod file
         uses: actions/setup-go@v6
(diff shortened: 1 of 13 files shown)


leokondrashov:
linters errors could be solved as here: https://github.com/vhive-serverless/vHive/commit/7139ed7f2f899e9b9de9e34c4776a6b9e6a4cad6, need to bump a config

16fb:
Okies, seem like the linter is functional now.

Should i aim to fix the linter errors in the code base (cmd and pkg)?
- One of the error is the `govet` linter not yet supporting type parameter inference for inline expansions. Placing explicit types in the function, gives an info popup in vscode for `unecessary type arguments`. Should the solution be to disable govet's inline checking?

The `Resolve Modules` test has outdated functions (which triggers a warning on future deprecation), and it seems to not be used later by the other tests `cmd` `pkg` `tools/plotter` which are manually defined in the workflow actions file.
- Should i aim to fix this warning also?

leokondrashov:
yes, please. updating go linter usually means that we need to update the codebase for the new rules.

the slices.contains is a weird one. are those reproduced with the config as in vhive as well? you can update the logic to fix this thing, but i'm fine with disabling the specific behavior check for govet (or making it non-error but a warning, as it should be).

the resolve module might not be needed, you can check what it does and delete it if it is so. i don't see that one in vhive, so it might be something legacy.

16fb:
The `govet` linter has been set to ignore "inline". Was not able to find a method to set only inline as warnings, so have set the inline check to be disabled for now.

Other linters errors, i have made the changes. However i have a questions with one of the recommended changes.
Linker: staticcheck, Rule: [QF1001](https://staticcheck.dev/docs/checks/#QF1001): could apply De Morgan's law.
Is this change ideal? or should it be disabled?. It works well for more complicated snippets, but some snippets can look like below:

Original
```
if !(successCount == 0 && failureCount == 1 && functionsInvoked == 1)
```

Recommended Fix
```
if successCount != 0 || failureCount != 1 || functionsInvoked != 1
```

Saw that developers from [another project](https://github.com/influxdata/telegraf/issues/16826) had the same question.

leokondrashov (approved):
looks good to me now

PR #784, opened 2026-08-04
Add zero-ing of invocations for Azure2021 Loader

16fb:
## Summary

After invocations read from csv, subtracts all timestamps by the smallest timestamp.
Forces the first invocation to start immediately. 

* After invocations are parsed but before function specifications are generated, `ZeroInvocations` finds the smallest invocation timestamp, and subtracts all timestamps by that invocation timestamp.
* Leaves `GenerateFunctionSpecification` `ParseCSVFile` mostly untouched.
* Minor changes in test functions to accommodate new expected output values.

none (N/A) 

* Earliest function invocation will fire immediately.


leokondrashov:
superceeded by #785

PR #785, opened 2026-08-11
Add zero-ing of Azure2021 filter function

16fb:
## Summary
* Zeros the resultant azure2021 trace to the smallest `start_timestamp` value.

* Add in `zero_timestamps` function in filter.py
* After keeping functions present in both sampled and original trace, subtracts the smallest `start_timestamp` value from both `start_timestamp` and `end_timestamp` columns.

none (N/A)

none (N/A)

### DIFF
--- a/docs/sampler.md
+++ b/docs/sampler.md
@@ -175,7 +175,7 @@ in [`plotTimeline/analysis.ipynb`](/tools/plotTimeline/analysis.ipynb)
 The sampling process of Azure2021 trace follows 3 steps: 
 1. Pre-process the original trace (`preprocess2021`) -> obtain clean trace in azure2019 format.
 2. Run the sampler on cleaned trace (`sampler`) -> obtain down-sampled trace, which contains a list of functions to keep in azure2019 format.
-3. Filter the original trace for functions in list (`filter`) -> obtain down-sampled and zero-ed trace in azure2021 format (per-invocation format).
+3. Filter the original trace for functions in list (`filter`) -> obtain down-sampled and timestamp normalized in azure2021 format (per-invocation format).
 
 Functionality of modules emulates the normal azure2019 modules, the differences is discussed below.
 
@@ -262,7 +262,7 @@ The preprocessing follows 3 steps:
 3. Transform the trace to azure2019 format. 
 
 ## Using IBM2026 Traces
-IBM2026 traces are supported by transforming it into Azure2021 format in `convertIBM2026`.
+IBM2026 traces are supported by transforming it into Azure2021 format in `preprocessIBM2026`.
 It is then treated as an Azure2021 trace in InVitro for Sampler and Loader. 
 
 ### Folder structure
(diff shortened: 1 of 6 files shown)


sampler/tests/filter_test.py
```diff
@@ -154,3 +154,56 @@ def test_same_app_different_functions(tmp_path):
...
+
+    # Read and compare output filtered_DF
+    filtered_df_path = out_dir / "SampledAzure2021.csv"
+    filtered_df = pd.read_csv(filtered_df_path)
+
+    expected_filtered_df = pd.DataFrame(
+        {
+            "app":           ["ac",  "ac"],
+            "func":          ["fa",  "fb"],
+            "end_timestamp": [20.0,  40.0],
```
leokondrashov: can you add the test that would end up with non-zero timestamp for the first invocation as we discussed? this should subtract 60s (the start_time), not necesserily the minimum of start timestamps

docs/sampler.md
```diff
@@ -175,7 +175,7 @@ in [`plotTimeline/analysis.ipynb`](/tools/plotTimeline/analysis.ipynb)
 The sampling process of Azure2021 trace follows 3 steps: 
 1. Pre-process the original trace (`preprocess2021`) -> obtain clean trace in azure2019 format.
 2. Run the sampler on cleaned trace (`sampler`) -> obtain down-sampled trace, which contains a list of functions to keep in azure2019 format.
-3. Filter the original trace for functions in list (`filter`) -> obtain down-sampled trace in azure2021 format (per-invocation format).
+3. Filter the original trace for functions in list (`filter`) -> obtain down-sampled and zero-ed trace in azure2021 format (per-invocation format).
```
leokondrashov: i would call it timestamp normalization, not zero-ed. that might be misunderstood

sampler/tests/filter_test.py
```diff
@@ -91,7 +91,7 @@ def test_azure2021_filter(tmp_path):
         {
             "app":           ["aa", "ab", "ac", "ac"],
             "func":          ["fa", "fb", "fd", "fd"],
-            "end_timestamp": [1.00, 10.0, 100.0, 300.0],
+            "end_timestamp": [ 1.0, 10.0, 100.0, 300.0], # All zeroed to start_time's Timedelta of 0
```
leokondrashov: so, unchanged, right? i would just remove the comment, not to confuse anyone

sampler/tests/filter_test.py
```diff
@@ -154,3 +154,56 @@ def test_same_app_different_functions(tmp_path):
...
 
+# - Test that output DF zeros to start_time
+def test_zero_to_start_time(tmp_path):
+
+    # Original DF
+    og_df = pd.DataFrame(
+        {
+            "app":           ["ac",  "ac"],
+            "func":          ["fa",  "fb"],
+            "end_timestamp": [80.0, 100.0], # 4 Minutes, 60-300 seconds
```
leokondrashov: the comment is weird, i guess it is copypasted from the other test with more requests

leokondrashov (commented):
looks good, a few rough small nitpicks here and there and good to go

sampler/tests/filter_test.py
16fb: Oops mb, didn't notice that the 2nd timestamp started at 0.

16fb:
Have incorporated the fixes you mentioned.

The Link check is returning 403 forbidden for a link to https://www.ibm.com/docs/en/cloud-private/3.1.2?topic=networking-kubernetes-network-model .

I think lychee has an .ignore/exclude file, should the link/domain be added to that?

leokondrashov:
if you follow the link, there is a banner at the top that there is newer version of the article, maybe they are responding with 403 because of that. can you try updating the link? if it doesn't help, try to find another article that covers the content. for sure there are some other posts about the pod cidrs that would be responding normally.

leokondrashov (dismissed):
fix the link, otherwise ready to merge

16fb:
Changed CIDR explanation link to https://www.redhat.com/en/blog/cni-kubernetes

leokondrashov (approved):

PR #796, opened 2026-09-13
Add supported trace documentation

16fb:
## Summary

Add documentation on supported traces.

* Offline copy of diagrams saved as .pdf files
* Main document `supporting_traces.md`

none (N/A)

none (N/A)


docs/supported_traces.md
```diff
@@ -0,0 +1,91 @@
...
+
+Loader and Sampler support for other formats is achieved by transforming them into either of the 2 main formats.
+- Huawei2023 -> Cleaned Azure2019
+- IBM2026 -> Azure2021
+
+## Diagrams
+The diagrams can be viewed through the following methods:
+- Online through this [public link](https://viewer.diagrams.net/?tags=%7B%7D&lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=Invitro%20Trace%20Format%20Table.drawio&dark=auto#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1OfL72TZTf9I1kZA0CnPJlyEsK-hys4Ah%26export%3Ddownload). 
+- Reopened in the [draw.io browser](https://app.diagrams.net/) by importing the workspace file at `./supported_traces/trace_format_diagrams.drawio`.
+- Viewing the [pdf directly](./supported_traces/trace_format_diagrams.pdf).
```
leokondrashov, line 29: i would just insert png/pdf in the markdown instead. the gdrive link might go stale at some point. still can upload the drawio as a file here, but as a reference or when we need to update it.

docs/supported_traces.md
```diff
@@ -0,0 +1,91 @@
+## Context
+This document serves to explain how different trace types are organised and used within InVitro.
+
+As InVitro supports more trace types, developers are required to keep track of multiple diverse trace schema, making development tedious and error-prone. Furthermore, users of InVitro need to understand how we preprocess and convert a given trace, particularly when information is removed or added.
+
+While this document aims to be as complete as possible, the most up-to-date information should be referenced from the direct documentation files (`loader.md` and `sampler.md`) or read from source directly.
```
leokondrashov, line 6: it sounds a bit weird talking about developers in third person. i think the second paragraph is rather pointless, so can just drop it.

docs/supported_traces/trace_format_diagrams.drawio
leokondrashov, line 1: for other docs, we just put the figures in the docs folder, but we can create a subfolder just for figures. calling it supported trace is weird, tho, just figures would be fine.

docs/supported_traces.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Diagrams
+The diagrams can be viewed through the following methods:
+- Online through this [public link](https://viewer.diagrams.net/?tags=%7B%7D&lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=Invitro%20Trace%20Format%20Table.drawio&dark=auto#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1OfL72TZTf9I1kZA0CnPJlyEsK-hys4Ah%26export%3Ddownload). 
+- Reopened in the [draw.io browser](https://app.diagrams.net/) by importing the workspace file at `./supported_traces/trace_format_diagrams.drawio`.
+- Viewing the [pdf directly](./supported_traces/trace_format_diagrams.pdf).
+
+### Summary Diagrams
+Simplified Schema Table
+- Brief summary of trace schema
```
leokondrashov, line 33: so, where is the table? please embed it to the markdown

docs/supported_traces/trace_format_diagrams.pdf
leokondrashov, line 1: this file contains a lot of unstructured information, can you move the most of it into the markdown instead? it's better to leave diagrams in figures (one per figure) or make them mermaid (depends on the way it would render it, there is a probability of it not looking nice). all the tables are better embedded as a mardown table, rather than inside pdf.

leokondrashov, line 1: also, i don't think there is a point in duplicating the content from the original repos that much, maybe just leave the description which kind of data exists in the origin but we drop, also better to move to the markdown and structure every trace subsection like processing pipeline (maybe with example commands), used fields, added values (like the 200MB of memory), ignored fields with short description. this would overlap with the table, so the table would act as a summary and the rest is the details.

leokondrashov (commented):
overall, good stuff, but needs a bit more structuring to make it easier to parse. mostly, need to move the information from the pdf into markdown and embed the figures, so there is no need to jump between markdown file and pdf.

